PR #1, opened 2020-02-20
Generate sdk code and make serverAPI private

sam-nguyen7:
Initial commit.

Add logic to make server API private.


sam-nguyen7:
/hold
need to refactor

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	})
+	if err != nil {
+		reqLogger.Error(err, "Failed to get AWS client")
+		return reconcile.Result{}, err
+	}
+
+	// To make API server non-public
+	// 1. Delete ext NLB for port 6443/TCP
+	// 2. Update api.<cluster-domain> DNS record to point to internal NLB
+	if instance.Spec.DefaultAPIServerIngress.Listening == cloudingressv1alpha1.Internal {
```
lisa: I'd like to see a comment about what the goal of this comparison is for. In the line of "If <x>, then we have to <y>" where x is a statement of customer request and y is the high level action we're taking. For example:

> // If the CR is wanting the default API server to be internal facing only, we need to: <list of steps in sequence to accomplish the goal>

The reason is that it helps folks to get inside your head and set some expectations for the block of code to come. People will mentally be checking things off the <list of steps> as they read through the code.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		// get a list of load balancers in use
```
lisa: Is the list specific to our cluster, or all of them in the account? (We can probably narrow it to the ones used by our cluster by tag)

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+			return reconcile.Result{}, err
+		}
+		
+		// get the internet-facing NLB by scheme and delete
+		var internalNLBDNS *string
+		var internalCanonicalHostedZoneID *string
+		for _, loadBalancers := range descLBOutput.LoadBalancers{
+			external := "internet-facing"
+			scheme := *loadBalancers.Scheme
+			if scheme == external {
```
lisa: Can you compare `if scheme == "internet-facing"`?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		descLBOutput, err := awsClient.DescribeLoadBalancersV2(&descLBInput)
+		if err != nil {
+			log.Error(err, "Error getting list of NLBs")
+			return reconcile.Result{}, err
+		}
+		
+		// get the internet-facing NLB by scheme and delete
+		var internalNLBDNS *string
+		var internalCanonicalHostedZoneID *string
+		for _, loadBalancers := range descLBOutput.LoadBalancers{
```
lisa: If you're looping over a slice of load balancers, the variable you use inside the loop to talk about one of those load balancers should be singular, eg `loadBalancer`, to remind folks its contents are for a single load balancer.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+			external := "internet-facing"
+			scheme := *loadBalancers.Scheme
+			if scheme == external {
+				//need to get Amazon Resource Name (ARN) of the load balancer to pass in input
+				extARN := loadBalancers.LoadBalancerArn
+				deleteLBInput := elbv2.DeleteLoadBalancerInput{extARN}
```
lisa: Can you combine these? So it's `elbv2.DeleteLoadBalancerInput{loadBalancers.LoadBalancerArn}`?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+				internalNLBDNS = loadBalancers.DNSName
+			}
+		}
+
+		// change Alias of external LB in public hosted zone to internal LB
+
+		dnsSpec := configv1.DNSSpec{}
+		domainName := dnsSpec.BaseDomain // in form of ```samn-test.j5u3.s1.devshift.org```
+
+		// get index of first period
```
lisa: Why is it important to get the index of the first period? What does it represent?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		perIndex := strings.Index(domainName, ".")		
+		// get pubDomainName in form of ```j5u3.s1.devshift.org```
+		pubDomainName := domainName[perIndex+1:len(domainName)]
```
lisa: I think it's okay to combine  these? `pubDomainName := domainName[strings.Index(domainName, ".")+1:len(domainName)]` doesn't seem too unwieldy? What do you think?  If we have a good comment explaining why it's important to do this operation on the cluster domain name. (eg, "in order to update DNS (this would be one of the "list of steps" earlier), we need the route53 zone name, which happens to be the cluster domain minus the name of the cluster" or something like that).

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		lookup := route53.ListHostedZonesByNameInput{
+			DNSName: aws.String(pubDomainName),
+		}
+		listHostedZones, err := awsClient.ListHostedZonesByName(&lookup)
+		var publicHostedZoneID string
+		for _, zone := range listHostedZones.HostedZones {
+			if zone.Config.PrivateZone == false {
+				// get public hostedzoneID
+				zoneID := *zone.Id
+				slashIndex := strings.LastIndex(zoneID)
```
lisa: Is this a syntax error? `strings.LastIndex` wants two parameters?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+				publicHostedZoneID = zoneID[slashIndex+1:len(zoneID)]
```
lisa: This line could probably be combined with the two previous, and expand the comment on line 181 to say why we need this.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+		pubDomainName := domainName[perIndex+1:len(domainName)]
+
+		// list hostedzones and get public hostedzoneID
+		lookup := route53.ListHostedZonesByNameInput{
+			DNSName: aws.String(pubDomainName),
+		}
+		listHostedZones, err := awsClient.ListHostedZonesByName(&lookup)
+		var publicHostedZoneID string
+		for _, zone := range listHostedZones.HostedZones {
+			if zone.Config.PrivateZone == false {
```
lisa: What happens if there's more than 1 `zone` that meets this criteria?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: This is a great point.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: yes

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: In theory an account shouldn't have more resources than what the cluster is using

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: agree

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: yes

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: explained in comment below

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: I agree. As long as the comment is good I don't see an issue with a one-liner.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: yes I corrected this and moved to route53 helper

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: that's too long in my opinion, we would have
`publicHostedZoneID = aws.StringValue(zone.Id)[strings.LastIndex(zoneID, "/")+1 : len(aws.StringValue(zone.Id))]`

sam-nguyen7: agree that we need better comment however

lisa: oh wow, that is a bit lengthy. 👍 to keeping it split for now, and giving the comment some attention.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: we and add a check that if `zone.Config.PrivateZone == false || zone.Name == aws.String(publicDomainName)` since this is the one we really care about

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// TODO(user): Modify this to be the types you create that are owned by the primary resource
+	// Watch for changes to secondary resource Pods and requeue the owner PublishingStrategy
+	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &cloudingressv1alpha1.PublishingStrategy{},
+	})
+	if err != nil {
+		return err
+	}
```
lisa: Missed this the first time around. I don't think we need to watch pods that we own?

pkg/apis/cloudingress/v1alpha1/publishingstrategy_types.go
```diff
@@ -0,0 +1,81 @@
...
+	Listening Listening `json:"listening,omitempty"`
+	// Default defines default value of ingress when cluster installs
+	Default       bool                   `json:"default"`
+	DNSName       string                 `json:"dnsName"`
+	Certificate   corev1.SecretReference `json:"certificate"`
+	RouteSelector metav1.LabelSelector   `json:"routeSelector"`
+}
+
+// Listening defines internal or external api and ingress
+type Listening string
```
fahlmant: Is this type used outside of the const and the two struct instances definitons? I'm questioning how useful this type alias is

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: We do not. Will remove.

pkg/apis/cloudingress/v1alpha1/publishingstrategy_types.go
sam-nguyen7: could just be used as const strings

pkg/awsclient/route53_helper.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+
+	// get public hosted zone ID needed to changeResourceRecordSets
+	var publicHostedZoneID string
+	for _, zone := range listHostedZones.HostedZones {
+		if zone.Config.PrivateZone == aws.Bool(false) || zone.Name == aws.String(clusterDomain) {
+			// In order to get the publicHostedZoneID we need to get
+			// the HostedZone.Id object which is in the form of "/hostedzone/Z1P3C0HZA40C0N"
+			// Since we only care about the ID number, we take index of the last "/" char and parse right
```
lisa: 👍 great comment

pkg/awsclient/ec2_helper.go
```diff
@@ -0,0 +1,188 @@
...
+func (c *awsClient) findSecurityGroupByName(name string) (*ec2.SecurityGroup, error) {
+	i := &ec2.DescribeSecurityGroupsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("group-name"),
+				Values: aws.StringSlice([]string{name}),
+			},
+		},
+	}
+	o, err := c.DescribeSecurityGroups(i)
```
fahlmant: Does this return an err if the group doesn't exist? If so, then line 174 would never be called.

lisa: It does not return an error.

pkg/awsclient/elb_helper.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return &AWSLoadBalancer{}, err
+	}
+	fmt.Printf("    * Adding health check (HTTP:6443/)\n")
+	err = c.addHealthCheck(elbName, "HTTP", "/", 6443)
+	if err != nil {
+		return &AWSLoadBalancer{}, err
+	}
+	// Caller will need the DNS name and Zone ID for the ELB (for route53) so let's make a handy object to return, using the
+	_, awsELBObj, err := c.DoesELBExist(elbName)
```
fahlmant: Recommendation: Occasionally objects take a while to become available via the API. So if this describe is called immediately after the create, it may not come back as available. In the account operator, we've implemented a retry so that the API inconsistenciesfailui don't lead to accidental failure prematurely.

pkg/awsclient/elb_helper.go
```diff
@@ -0,0 +1,170 @@
...
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case elb.ErrCodeAccessPointNotFoundException:
+				return false, &AWSLoadBalancer{}, nil
+			default:
+				return false, &AWSLoadBalancer{}, err
+			}
+		}
+	}
+	return true, &AWSLoadBalancer{ELBName: elbName, DNSName: *res.LoadBalancerDescriptions[0].DNSName, DNSZoneId: *res.LoadBalancerDescriptions[0].CanonicalHostedZoneNameID}, nil
```
fahlmant: why not just return the elb in res from line 143 directly?

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -0,0 +1,153 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Define a new Pod object
+	pod := newPodForCR(instance)
```
fahlmant: I'd recommend removing from here down, as you'll accidentally create pods if you test this controller.

pkg/awsclient/ec2_helper.go
fahlmant: Documentation on that?

pkg/awsclient/elb_helper.go
```diff
@@ -0,0 +1,170 @@
...
+				Protocol:         aws.String("tcp"),
+				LoadBalancerPort: aws.Int64(listenerPort),
+			},
+		},
+	}
+	_, err := c.CreateLoadBalancer(i)
+	if err != nil {
+		return &AWSLoadBalancer{}, err
+	}
+	fmt.Printf("    * Adding health check (HTTP:6443/)\n")
```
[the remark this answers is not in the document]
lisa: Whoops. Once it's merged, I'll take care to purge it when Sam and I end up merging our code here.

pkg/awsclient/elb_helper.go
lisa: The retry happens in the apischeme controller, and not here in the library. That code is to be committed.

pkg/awsclient/elb_helper.go
lisa: I don't want the controller(s) to deal in cloud-specific objects where possible. Clearly here the `AWS` gives it away, but in the future, this may change name or type (eg to interface).

pkg/awsclient/ec2_helper.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/elb"
+)
+
+// FIXME: Add tags
+
+// EnsureCIDRAccess ensures that for the given load balancer, the specified CIDR
+// blocks, and only those blocks may access it.
+// cidrBlocks always goes from 6443/TCP to 6443/TCP and is IPv4 only
```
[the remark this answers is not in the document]
lisa: Sure, that will be reflected once my wip branch is merged.

pkg/awsclient/ec2_helper.go
```diff
@@ -0,0 +1,188 @@
...
+	// to cidrBlocks, but that doesn't always hit the expected set, so
+	// this is a map to see if we have done just that. Any which are
+	// false were not processed.
+	seenExpectedRules := make(map[string]bool)
+	// init map
+	for _, cidrBlock := range cidrBlocks {
+		seenExpectedRules[cidrBlock] = false
+	}
+	// For each ingress rule for the security group,
+Outer:
```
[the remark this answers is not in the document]
lisa: That's how `go fmt` decided to format it 🤷‍♀ 

I agree that it looks visually strange, but that's what came out.

The purpose of the `Outer` label is to  skip processing the rule once we've encountered it in the inner loop (lines 52-60) because by that time, it's already been processed.

pkg/awsclient/ec2_helper.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	if err := c.addIngressRulesToSecurityGroup(securityGroup, rulesToAdd); err != nil {
+		return err
+	}
+	if err := c.removeIngressRulesFromSecurityGroup(securityGroup, rulesToRemove); err != nil {
+		return err
+	}
+	// Once the ingress rules are updated, attach the SecurityGroup to the load balancer
+	return c.setLoadBalancerSecurityGroup(loadBalancerName, securityGroup)
```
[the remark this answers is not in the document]
lisa: Good point. I'll add that as a todo item in my branch.

pkg/awsclient/ec2_helper.go
lisa: The behaviour in [the documentation for this method](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeSecurityGroups) does not indicate that any error is returned when there are no matches in the specified filters. Additionally, the [common errors](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#CommonErrors) for the library fail to indicate this occurrence. Granted, the absence of this documentation doesn't mean that it doesn't happen, but it is hard to prove a negative. ;)

As for anecdotal evidence, I have a small test program I used to back up my statement.

pkg/awsclient/elb_helper.go
lisa: There's another one on line 24! Both have been removed in my working copy that will be merged once Sam's code is merged.

pkg/awsclient/elb_helper.go
```diff
@@ -0,0 +1,221 @@
...
+	ExtScheme                    string
+	IntCanonicalHostedZoneNameID string
+	IntDNSName                   string
+	IntLoadBalancerArn           string
+	IntLoadBalancerName          string
+	IntScheme                    string
+}
+
+// ListAllNLBs uses the DescribeLoadBalancersV2 to get back a list of all Network Load Balancers
+func (c *awsClient) ListAllNLBs() (LoadBalancerInfo, error) {
```
lisa: I really like the idea of combining the `elb` and `elbv2` calls into one `elb_helper.go`. 👍

pkg/awsclient/elb_helper.go
```diff
@@ -0,0 +1,221 @@
...
+			LBInfo.IntDNSName = aws.StringValue(loadBalancer.DNSName)
+			LBInfo.IntLoadBalancerArn = aws.StringValue(loadBalancer.LoadBalancerArn)
+			LBInfo.IntLoadBalancerName = aws.StringValue(loadBalancer.LoadBalancerName)
+		}
+	}
+	return LBInfo, nil
+}
+
+// DeleteExternalLoadBalancer takes in the external LB arn and deletes the entire LB
+func (c *awsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error {
```
lisa: For people wondering what happens if the load balancer is already deleted:

> If the load balancer does not exist or has already been deleted, the call succeeds.

[source](https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#ELBV2.DeleteLoadBalancer)

pkg/awsclient/elb_helper.go
```diff
@@ -0,0 +1,221 @@
...
+	ExtDNSName                   string
+	ExtLoadBalancerArn           string
+	ExtLoadBalancerName          string
+	ExtScheme                    string
+	IntCanonicalHostedZoneNameID string
+	IntDNSName                   string
+	IntLoadBalancerArn           string
+	IntLoadBalancerName          string
+	IntScheme                    string
+}
```
lisa: How about treating this like a list so that in lines 182-194 it's dealing with a list of all the external load balancers, which is returned to the caller. It might be that we need more information about the load balancer, such as the ports, or whatever else to be sure that we eventually delete the correct one, in case there's more than the one external nlb. (I also changed the name just for clarity in the snippet)

```go
// LoadBalancerV2 is a list of non-classic AWS ELBs
type LoadBalancerV2 struct {
  HostedZoneNameId string
  DNSName          string
  Arn              string
  Name             string
  Scheme           string
// ...
}
```

With this kind of structure, there's less duplication in the structure.

Later, the loop could look like:

```go
loadBalancers := make([]LoadBalancersV2, 0)
for _, loadBalancer := range output.LoadBalancers {
	loadBalancers = append(loadBalancers, LoadBalancersV2{
		HostedZoneNameId: loadBalancer.CanonicalHostedZoneId,
		DNSName:          loadBalancer.DNSName,
		Arn:              loadBalancer.LoadBalancerArn,
		Name:             loadBalancer.LoadBalancerName,
		Scheme:           loadBalancer.Scheme,
	})
}

return loadBalancers, nil
```

Since this method is currently mainly used to delete the external NLB, that caller could range through the list to find the appropriate one and perform the deletion before `break`ing out of the loop.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -0,0 +1,151 @@
...
+		// which happens to be the domainName minus the name of the cluster
+		// Since there are NO object on cluster with just clusterName,
+		// we will index the first period and parse right
+		pubDomainName := domainName[strings.Index(domainName, ".")+1:len(domainName)]	// pubDomainName in form of ```j5u3.s1.devshift.org```
+		apiDNSName := "api." + domainName + "."
+		comment := "Update api.<clusterName> alias to internal NLB"
+
+		// upsert resource record to change api.<clusterName> from external NLB to internal NLB
+		awsClient.UpsertCNAME(pubDomainName, loadBalancerInfo.IntDNSName, loadBalancerInfo.IntCanonicalHostedZoneNameID, apiDNSName, comment, false)
+		return reconcile.Result{}, nil
```
lisa: Error checking?

pkg/awsclient/elb_helper.go
sam-nguyen7: That feels very AWS specific to me. Since we don't know what other cloud api looks like does it make more sense to refactor later on?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: will add

pkg/awsclient/elb_helper.go
lisa: I think it makes sense to handle it now as a code clarity change. Remember, this is the aws client helper, so it makes sense to be aws-specific here. Once we get to the multi-cloud world we already have to refactor AWS-specific parts out, so we can take that time do it. Here, in our AWS library, we can be as AWS-specific as we like.

sam-nguyen7: we use this method to both delete the external NLB but also get some information about the internal NLB for `upsertCNAME` func. if we use this method and I imagine in our controller we would have to put AWS specific logic like `scheme == internal`. I was trying to avoid having to do this

lisa: That's true. What if the controller is calling something like `DeleteExternalLoadBalancer` and passes in what the name of that external load balancer is? It could come from the `Machine` object. In this way the controller doesn't need to know what kind of load balancers are under the hood?

deploy/crds/cloudingress.managed.openshift.io_publishingstrategies_crd.yaml
```diff
@@ -0,0 +1,137 @@
...
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: PublishingStrategySpec defines the desired state of PublishingStrategy
+          properties:
+            applicationIngress:
+              description: ApplicationIngresws defines whether application ingress
```
csheremeta: s/ApplicationIngresws/ApplicationIngress

pkg/awsclient/client.go
```diff
@@ -0,0 +1,258 @@
...
+func (c *awsClient) DeregisterInstancesFromLoadBalancer(i *elb.DeregisterInstancesFromLoadBalancerInput) (*elb.DeregisterInstancesFromLoadBalancerOutput, error) {
+	return c.elbClient.DeregisterInstancesFromLoadBalancer(i)
+}
+func (c *awsClient) DescribeLoadBalancers(i *elb.DescribeLoadBalancersInput) (*elb.DescribeLoadBalancersOutput, error) {
+	return c.elbClient.DescribeLoadBalancers(i)
+}
+
+func (c *awsClient) DescribeLoadBalancersV2(i *elbv2.DescribeLoadBalancersInput) (*elbv2.DescribeLoadBalancersOutput, error) {
+	return c.elbv2Client.DescribeLoadBalancers(i)
+}
```
lisa: SUPER unfortunate that these have the same upstream name. :( 👍 on your name for the elbv2 one.

deploy/crds/cloudingress.managed.openshift.io_publishingstrategies_crd.yaml
```diff
@@ -0,0 +1,137 @@
...
+                            operator:
+                              description: operator represents a key's relationship
+                                to a set of values. Valid operators are In, NotIn,
+                                Exists and DoesNotExist.
+                              type: string
+                            values:
+                              description: values is an array of string values. If
+                                the operator is In or NotIn, the values array must
+                                be non-empty. If the operator is Exists or DoesNotExist,
+                                the values array must be empty. This array is replaced
```
csheremeta: s/must/may ?

lisa:
@sam-nguyen7 Aside from @csheremeta's comments, I think this is good to start testing. I'm sure more will turn up, and I can see some things that give pause, but let's get to those Later ™️ .

deploy/crds/cloudingress.managed.openshift.io_publishingstrategies_crd.yaml
sam-nguyen7: this is auto-generated from spec file, does it make sense to change it? @csheremeta @lisa

csheremeta: hm, I'm not sure. @lisa do you have an opinion?

lisa: I'd regenerate. It's important to do without typos since they will be used to auto-generate docs.

sam-nguyen7: Will do.

PR #6, opened 2020-03-02
Add machine controller to watch Machine objects for

csheremeta:
when status.providerStatus.instanceID changes, remove old instanceID from LB, add new one to LB


csheremeta:
https://issues.redhat.com/browse/OSD-2831

sam-nguyen7:
/lgtm
logic makes sense. I'm good with merging this in but we should create a card to track testing

PR #7, opened 2020-03-05
Make server API public

sam-nguyen7:
Steps taken to make server API public:
    1. create a new external NLB
    2. create target group
    3. get master instances and retrieve AZs and IPs of each
    4. register these AZs and IPs with created target group
    5. create the listener to connect NLB to target group

EDIT:
New procedure to make server API public:
    1. create new external NLB
    2. look for old/existing TargetGroup
    3. create listener to link new external NLB to said targetGroup
    4. update route53 api.<clustername> alias to be the new NLB

EDIT 2:
    I have tested this manually on a stage cluster. It works but there are a lot of error handling that I have not implemented. This is the minimal solution for now. I will document all error and edge handling cases and revisit them in the next iteration


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -160,23 +180,142 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		// upsert resource record to change api.<clusterName> from internal NLB to external NLB
+		comment := "Update api.<clusterName> alias to external NLB"
+		err = awsClient.UpsertARecord(pubDomainName+".", newNLBs[0].DNSName, newNLBs[0].CanonicalHostedZoneNameID, apiDNSName, comment, false)
+		if err != nil {
+			log.Error(err, "Error updating api.<clusterName> alias to internal NLB")
+			return reconcile.Result{}, err
+		}
+		log.Info(comment + " successful")
+
+		// update route53 api.<cluster-name> with external NLB
```
rafael-azevedo: Why commented out code here as well ?

pkg/controller/utils/clusterinfo.go
```diff
@@ -127,6 +127,50 @@ func GetClusterMasterInstances(kclient client.Client) ([]string, error) {
...
+	listOptions := []client.ListOption{
+		client.InNamespace("openshift-machine-api"),
+		client.MatchingLabels{masterMachineLabel: "master"},
+	}
+	err := kclient.List(context.TODO(), machineList, listOptions...)
+	if err != nil {
+		return []MasterInstance{}, err
+	}
+
+	masterInstances := make([]MasterInstance, 0)
```
rafael-azevedo: you should initialize this as well like it was mentioned in the function of the awsclient

pkg/awsclient/elb_helper.go
```diff
@@ -212,6 +215,123 @@ func (c *AwsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error
...
+		HealthyThresholdCount:      aws.Int64(2),
+		UnhealthyThresholdCount:    aws.Int64(2),
+	}
+
+	result, err := c.CreateTargetGroupV2(i)
+	if err != nil {
+		return "", err
+	}
+
+	return aws.StringValue(result.TargetGroups[0].TargetGroupArn), nil
```
rafael-azevedo: @sam-nguyen7 Also looking at the api client in this case you should just return an error after creating. Since the result contains an array and you don't know explicitly which index the of the result is the data you want. After you create you should do a get call with the nlbName to get the data you want.

rafael-azevedo: There are a few functions in the awsclient library where I see this occuring

pkg/awsclient/elb_helper.go
```diff
@@ -212,6 +215,123 @@ func (c *AwsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error
...
+	}
+
+	result, err := c.CreateLoadBalancerV2(i)
+	if err != nil {
+		return []LoadBalancerV2{}, err
+	}
+
+	// there should only be 1 NLB made, but since CreateLoadBalancerOutput takes in slice
+	// we return it as slice
+	loadBalancers := make([]LoadBalancerV2, 0)
```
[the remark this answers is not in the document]
sam-nguyen7: noted

pkg/awsclient/elb_helper.go
```diff
@@ -212,6 +215,123 @@ func (c *AwsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error
...
+		Names: []*string{
+			aws.String(targetGroupName),
+		},
+	}
+
+	result, err := c.DescribeTargetGroupsV2(i)
+	if err != nil {
+		return "", err
+	}
+	return aws.StringValue(result.TargetGroups[0].TargetGroupArn), nil
```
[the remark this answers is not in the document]
sam-nguyen7: noted

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -160,23 +180,142 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+			return reconcile.Result{}, err
 		}
+		log.Info(comment + " successful")
 		return reconcile.Result{}, nil
 	}
 
+	// if CR is wanted the default server API to be internet-facing, we
+	// create the external NLB for port 6443/TCP and add api.<cluster-name> DNS record to point to external NLB
+	if instance.Spec.DefaultAPIServerIngress.Listening == cloudingressv1alpha1.External {
+		// get a list of all non-classic ELBs
```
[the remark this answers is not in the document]
sam-nguyen7: code commented out is for making a new targetGroup and listener etc. Since we have changed some of our logic this may not be needed, but we MIGHT need it later on. I would prefer to keep them around in case we need them later

pkg/awsclient/client.go
```diff
@@ -212,22 +221,40 @@ func (c *AwsClient) DescribeLoadBalancers(i *elb.DescribeLoadBalancersInput) (*e
...
 
+func (c *AwsClient) DescribeLoadBalancersV2(i *elbv2.DescribeLoadBalancersInput) (*elbv2.DescribeLoadBalancersOutput, error) {
+	return c.elbv2Client.DescribeLoadBalancers(i)
+}
+
+func (c *AwsClient) DeleteLoadBalancerV2(i *elbv2.DeleteLoadBalancerInput) (*elbv2.DeleteLoadBalancerOutput, error) {
+	return c.elbv2Client.DeleteLoadBalancer(i)
+}
+
+func (c *AwsClient) CreateLoadBalancerV2(i *elbv2.CreateLoadBalancerInput) (*elbv2.CreateLoadBalancerOutput, error) {
```
rogbas: Why are we adding this V2 suffix? Is it to avoid collisions with the version that Lisa is working on?

pkg/awsclient/elb_helper.go
```diff
@@ -212,6 +215,123 @@ func (c *AwsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error
...
+		})
+	}
+	return loadBalancers, nil
+}
+
+// CreateExternalNLBTargetGroup creates the external NLB target group and returns the targetGroupArn
+func (c *AwsClient) CreateExternalNLBTargetGroup(nlbName, vpcID string) (string, error) {
+	i := &elbv2.CreateTargetGroupInput{
+		Name:                       aws.String(nlbName),
+		Port:                       aws.Int64(6443),
```
rogbas: Probably we should consider transforming Port and Protocols into constants, and use them here.

pkg/awsclient/elb_helper.go
```diff
@@ -212,6 +215,123 @@ func (c *AwsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error
...
+		Protocol:                   aws.String("TCP"),
+		TargetType:                 aws.String("ip"),
+		VpcId:                      aws.String(vpcID),
+		HealthCheckPath:            aws.String("/readyz"),
```
rogbas: Same thing for the healthcheck data: path, port, protocol and intervals/thresholds, we should have constants for that

pkg/awsclient/elb_helper.go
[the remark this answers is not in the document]
rogbas: Is it possible for the TargetGroups slice to be empty? If yes this could crash when we access `TargetGroupArn` in an empty array. We would need to at least check for slice length before accessing it.

pkg/awsclient/elb_helper.go
```diff
@@ -212,6 +215,123 @@ func (c *AwsClient) DeleteExternalLoadBalancer(extLoadBalancerArn string) error
...
+		return "", err
+	}
+
+	return aws.StringValue(result.TargetGroups[0].TargetGroupArn), nil
+}
+
+// RegisterMasterNodeAZsandIPs register master node IPs and AZs
+// TODO: need to test function
+// This will take awhile before healthcheck is happy
+func (c *AwsClient) RegisterMasterNodeAZsandIPs(targetGroupArn string, mi []utils.MasterInstance) error {
```
rogbas: The call to this function is commented out in the Reconcile loop. Are we expected to uncomment that?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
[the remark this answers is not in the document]
rogbas: Is this supposed to be a WIP PR? Are we going to enable that part in a follow-up PR?

pkg/awsclient/client.go
fahlmant: There are 2 versions of the elb API, but the function names are the same from the aws sdk

pkg/awsclient/elb_helper.go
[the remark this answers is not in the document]
sam-nguyen7: yes it is possible. I would like to implement a better way to handle this instead of indexing the slice. For now I will put in check for slice length but will revisit this in the next iteration

sam-nguyen7: noted

pkg/awsclient/client.go
sam-nguyen7: V2 is specifically for non-classic ELB functions. Yes it is to avoid confusion as classic elb could also have a function called CreateLoadBalancer

pkg/awsclient/elb_helper.go
sam-nguyen7: noted. Since this is not being called in controller I will revisit.

pkg/awsclient/elb_helper.go
sam-nguyen7: noted. Since this is not being called in controller I will revisit.

pkg/awsclient/elb_helper.go
sam-nguyen7: in future iteration I believe we will. Discussion on-going in slack

pkg/controller/utils/clusterinfo.go
```diff
@@ -127,6 +127,50 @@ func GetClusterMasterInstances(kclient client.Client) ([]string, error) {
...
+		}
+		awsconfig := &awsproviderapi.AWSMachineProviderConfig{}
+		err = codec.DecodeProviderSpec(&machineList.Items[0].Spec.ProviderSpec, awsconfig)
+		if err != nil {
+			return masterInstances, err
+		}
+		az := awsconfig.Placement.AvailabilityZone
+
+		// get the IP address from Master object's status
+		ip := mi.Status.Addresses[0].Address
```
rogbas: It seems a little dangerous to me to just get the first element of the Addresses slice. Each machine will have 3 diff types of address (as example below). We're safer creating a quick helper function that will return the Address from the item that has `type == InternalIP`

Sample from Machine Status:
```yaml
status:
  addresses:
  - address: 10.0.138.191
    type: InternalIP
  - address: ip-10-0-138-191.ec2.internal
    type: InternalDNS
  - address: ip-10-0-138-191.ec2.internal
    type: Hostname
  lastUpdated: "2020-03-10T17:34:43Z"

```

rogbas: They could be in a diff order, hence my concern.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -131,13 +133,30 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	domainName, err := utils.GetClusterBaseDomain(r.client) // in form of ```samn-test.j5u3.s1.devshift.org```
+	if err != nil {
```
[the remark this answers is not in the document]
sam-nguyen7: I think GetClusterBaseDomain function should return error if err != nil . Will change it.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -160,23 +180,147 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		log.Info(fmt.Sprintf("subnet IDs : %v ", subnetIDs))
+		newNLBs, err := awsClient.CreateNetworkLoadBalancer(extNLBName, "internet-facing", subnetIDs[0]) // subnetID can be found by utils.GetMasterNodeSubnets(r.client)
+		if err != nil {
+			log.Error(err, "couldn't create external NLB")
+			return reconcile.Result{}, err
+		}
+		log.Info(fmt.Sprintf("new external NLB: %v", newNLBs))
+
+		if len(newNLBs) != 1 {
+			log.Error(err, "more than one NLB detected. Error out")
```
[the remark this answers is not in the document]
sam-nguyen7: change

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -160,23 +180,147 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		// upsert resource record to change api.<clusterName> from internal NLB to external NLB
+		comment := "Update api.<clusterName> alias to external NLB"
+		err = awsClient.UpsertARecord(pubDomainName+".", newNLBs[0].DNSName, newNLBs[0].CanonicalHostedZoneNameID, apiDNSName, comment, false)
+		if err != nil {
+			log.Error(err, "Error updating api.<clusterName> alias to internal NLB")
+			return reconcile.Result{}, err
+		}
+		log.Info(fmt.Sprintf("%s successful ", comment))
+
+		// update route53 api.<cluster-name> with external NLB
```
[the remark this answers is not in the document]
sam-nguyen7: yes

pkg/controller/utils/clusterinfo.go
sam-nguyen7: removed func

pkg/controller/publishingstrategy/publishingstrategy_controller.go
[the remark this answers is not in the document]
sam-nguyen7: issue created to track this.

rogbas:
After the latest changes it LGTM
/lgtm

PR #15, opened 2020-03-11
Logic for handling data ingresses

sam-nguyen7:
Implements the logic for our CR specifically publishingStrategy.Spec.applicationIngresses.

Goes through the []applicationIngresses and compare them with ingresscontroller objects on the cluster. Delete and create a new ingresscontroller if needed. Create a new ingresscontroller if needed. Exit reconcile loop if nothing is needed.


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+	for _, publishingStrategyIngress := range instance.Spec.ApplicationIngress {
+		for _, ingressController := range ingressControllerList.Items {
+			if doesIngressMatch(&publishingStrategyIngress, &ingressController) {
+				log.Info("match, move on to next cycle")
+				continue
+			} else {
+				// create a LocalObjectReference and fill it out secret passed in by OCM
+				newCertificate := &corev1.LocalObjectReference{
+					Name: publishingStrategyIngress.Certificate.Name,
+				}
```
rafael-azevedo: If you do !doesIngressMatch() you can clean up the logic alot.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+						}
+						err = r.client.Create(context.TODO(), newIngressController)
+						if err != nil {
+							log.Error(err, "failed to create new ingresscontroller CR")
+							return reconcile.Result{}, err
+						}
+						log.Info(fmt.Sprintf("succesfully create new ingresscontroller %s", newIngressControllerName))
+					}
+				}
+			}
```
rafael-azevedo: You should refactor this into at least two other function calls to make the logic easier to understand. 

Also when you start to use Else in go look back at your logic there is probably a simpler way to do it.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+	if string(publishingStrategyIngress.Listening) != string(ingressController.Spec.EndpointPublishingStrategy.Type) {
+		return false
+	}
+	if publishingStrategyIngress.DNSName != ingressController.Spec.Domain {
+		return false
+	}
+	if publishingStrategyIngress.Certificate.Name != ingressController.Spec.DefaultCertificate.Name {
+		if publishingStrategyIngress.Certificate.Namespace != ingressController.Namespace {
+			return false
+		}
```
rafael-azevedo: This logic should probably not be nested in this if loop. The fill loop will return false regardless of what happens in this if block.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+					}
+					err = r.client.Create(context.TODO(), newIngressControllerCR)
+					if err != nil {
+						log.Error(err, "failed to create new ingresscontroller CR")
+						return reconcile.Result{}, err
+					}
+					log.Info("succesfully created new ingresscontroller default")
+				}
+				// if the ingresscontroller name is not default
+				if ingressController.Name != "default" {
```
rogbas: This `if` could be transformed into an `else` of the `if ingressController.Name == "default"` statement above

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+					// if the content of this ingresscontroller matches
+					if doesIngressMatch(&publishingStrategyIngress, &ingressController) {
+						log.Info(fmt.Sprintf("%s already exists on cluster", ingressController.Name))
+						continue
+					} else { // if the ingresscontroller name is not default and the contents do not match
+						// create a new ingresscontroller
+						publishingStrategyIngressDNSName := publishingStrategyIngress.DNSName
+						firstPeriodIndex := strings.Index(publishingStrategyIngressDNSName, ".")
```
rogbas: I would transform that string manipulation operation into a function that returns the first part of the domain name

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+					err := r.client.Delete(context.TODO(), &ingressController)
+					if err != nil {
+						log.Error(err, "failed to delete existing ingresscontroller")
+						return reconcile.Result{}, err
+					}
+
+					// create a new ingresscontroller
+					newIngressControllerCR, err := newApplicationIngressControllerCR("default", string(publishingStrategyIngress.Listening), publishingStrategyIngress.DNSName, newCertificate)
+					if err != nil {
+						log.Error(err, "failed to generate data for new ingresscontroller CR")
```
rogbas: Here we find an error but we don't return and still continue to next step potentially using a bogus `newIngressControllerCR` variable

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -261,5 +266,138 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+						log.Info(fmt.Sprintf("%s already exists on cluster", ingressController.Name))
+						continue
+					} else { // if the ingresscontroller name is not default and the contents do not match
+						// create a new ingresscontroller
+						publishingStrategyIngressDNSName := publishingStrategyIngress.DNSName
+						firstPeriodIndex := strings.Index(publishingStrategyIngressDNSName, ".")
+						newIngressControllerName := publishingStrategyIngressDNSName[:firstPeriodIndex]
+						newIngressController, err := newApplicationIngressControllerCR(newIngressControllerName, string(publishingStrategyIngress.Listening), publishingStrategyIngress.DNSName, newCertificate)
+						if err != nil {
+							log.Error(err, "failed to generate data for new ingresscontroller CR")
```
rogbas: Same as above, here we find an error but we don't return and still continue to next step potentially using a bogus `newIngressController` variable

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: agree

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: agree

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: return reconcile.Result{}, err added

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: err added

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -284,9 +426,72 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+	if string(publishingStrategyIngress.Listening) != string(ingressController.Status.EndpointPublishingStrategy.LoadBalancer.Scope) {
+		return false
+	}
+	if publishingStrategyIngress.DNSName != ingressController.Spec.Domain {
+		return false
+	}
+	if publishingStrategyIngress.Certificate.Name != ingressController.Spec.DefaultCertificate.Name {
+		return false
+	}
+	if publishingStrategyIngress.Certificate.Namespace != ingressController.Namespace {
```
rogbas: Will this ever match? My understanding is `ingressController.Namespace` will always be = `openshift-ingress-operator`

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -100,6 +107,141 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		return reconcile.Result{}, err
+	}
+
+	// create temp list of applicationIngress
+	ingressNotOnClusterList := instance.Spec.ApplicationIngress
+	// loop through every applicationingress in publishing strategy and every ingresscontroller in cluster
+	for _, publishingStrategyIngress := range instance.Spec.ApplicationIngress {
+		for _, ingressController := range ingressControllerList.Items {
+			if !isOnCluster(&publishingStrategyIngress, &ingressController) {
+				ingressNotOnClusterList = append(ingressNotOnClusterList, publishingStrategyIngress)
```
rogbas: We're starting with `ingressNotOnClusterList` containing all the ApplicationIngress from the PublishingStrategy, then later we re-append into the same slice ingresses that don't match. Isn't that creating a bunch of duplicate ingresses in that slice?

rogbas: I thought we would in this first nested loop actually remove from `ingressNotOnClusterList` slice the ingresses that matched, resulting in a slice containing only Ingresses that are NOT in cluster (as the slice name states)

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: agreed. removing this condition

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: ah that is my mistake.`ingressNotOnClusterList` should be of type `[]cloudingressv1alpha1.ApplicationIngress` which is an empty slice. Then we append ones that don't match to the slice.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -100,6 +107,141 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+	// get a list of all ingress on the cluster
+	ingressControllerList := &operatorv1.IngressControllerList{}
+	listOptions := []client.ListOption{
+		client.InNamespace("openshift-ingress-operator"),
+	}
+	err = r.client.List(context.TODO(), ingressControllerList, listOptions...)
+	if err != nil {
+		log.Error(err, "Cannot get list of ingresscontroller")
+		return reconcile.Result{}, err
+	}
```
lisa: Note for future us:

In order to support this out-of-namespace access, this operator needs to have its `WATCH_NAMESPACE` envvar set to `` (empty string), or else the `List` operation will silently fail. (The reason it fails is because it will ignore the given namespace and look instead to the value of `WATCH_NAMESPACE`.)

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -100,6 +107,141 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+			for _, ingresscontroller := range ingressControllerList.Items {
+				if ingresscontroller.Name == defaultIngressName {
+					err := r.client.Delete(context.TODO(), &ingresscontroller)
+					if err != nil {
+						log.Error(err, "failed to delete existing ingresscontroller")
+						return reconcile.Result{}, err
+					}
+				}
+			}
+			log.Info(fmt.Sprintf("in default loop, appingress is: %v", appingress))
```
lisa: @sam-nguyen7 Was this meant to be deleted with cb8fd99?

sam-nguyen7: yes!

pkg/controller/publishingstrategy/publishingstrategy_controller.go
rogbas: /lgtm

rogbas: thanks for fixing that

Issue #13, opened 2020-03-11
Add gotest and gocheck make targets from gobuild


sam-nguyen7:
Per https://github.com/openshift/cloud-ingress-operator/pull/11 we could not get go list to work so have removed gotest and gocheck make targets from gobuild. Once we have testing we need to add these back

jewzaam:
Fixed in https://github.com/openshift/cloud-ingress-operator/pull/16

PR #16, opened 2020-03-11
Fix build/test in CI

cblecker:
This borrows some settings from the managed-velero-operator for building/testing the operator under go1.13. I believe this should fix https://github.com/openshift/release/pull/7597

/assign @jewzaam

### DIFF
--- a/standard.mk
+++ b/standard.mk
@@ -27,14 +27,13 @@ OPERATOR_DOCKERFILE ?=build/Dockerfile
 
 BINFILE=build/_output/bin/$(OPERATOR_NAME)
 MAINPACKAGE=./cmd/manager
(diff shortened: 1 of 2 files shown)


standard.mk
```diff
@@ -68,7 +67,7 @@ gocheck: ## Lint code
 
 .PHONY: gobuild
 gobuild: #gocheck gotest ## Build binary
```
jewzaam:
```suggestion
gobuild: gocheck gotest ## Build binary
```

Tried this locally, it works.  Recommend we go ahead and get it included and close https://github.com/openshift/cloud-ingress-operator/issues/13 and https://github.com/openshift/cloud-ingress-operator/pull/17

PR #22, opened 2020-03-13
Add logic to add/remove LBs from masters' ProviderSpecs

csheremeta:


pkg/controller/utils/clusterinfo.go
```diff
@@ -100,6 +100,21 @@ func GetClusterRegion(kclient client.Client) (string, error) {
 	return infra.Status.PlatformStatus.AWS.Region, nil
 }
 
+// GetMasterNodes returns a machineList object whose .Items can be iterated
+// over to perform actions on/with information from each master machine object
+func GetMasterNodes(kclient client.Client) (*machineapi.MachineList, error) {
```
mwoodson: I think this should be GetMasterMachineList

Nodes are a different thing in Openshift than machines.

pkg/controller/utils/clusterinfo.go
```diff
@@ -56,9 +56,8 @@ func GetClusterName(kclient client.Client) (string, error) {
 // }
 //
 func GetMasterNodeSubnets(kclient client.Client) (map[string]string, error) {
-	machineList := &machineapi.MachineList{}
 	subnets := make(map[string]string)
-	err := kclient.List(context.TODO(), machineList, client.InNamespace("openshift-machine-api"), client.MatchingLabels{masterMachineLabel: "master"})
+	machineList, err := GetMasterMachines(kclient)
```
mwoodson: can you redefine this like this?

PR #24, opened 2020-03-13
Fixing duplicate SSS

jewzaam:
https://issues.redhat.com/browse/OSD-3032

There were 2 SSS with the same name, so the last one always overwrote the first.


hack/templates/template.yaml
```diff
@@ -30,53 +30,4 @@ parameters:
...
-      kind: Subscription
-      metadata:
-        name: ${REPO_NAME}
-        namespace: openshift-${REPO_NAME}
-      spec:
-        channel: ${CHANNEL}
-        name: ${REPO_NAME}
-        source: ${REPO_NAME}-registry
-        sourceNamespace: openshift-${REPO_NAME}
+objects: []
```
cblecker: Is this right? It doesn't look right.

jewzaam:
/hold

ci.ext doesn't have `jq`.. heart breaking!

hack/templates/template.yaml
jewzaam: Script takes this and populates with child objects, which are SSS in our use case.  It's an array and we have nothing to start with, it is moved to separate yaml in deploy now (CatalogSource and Subscription).  To make sure python uses the right data type, `[]` so yes it's right.

jewzaam:
Removed bits for the PR check tweaks, pushed to another branch and I'll hack on it later.

PR #25, opened 2020-03-16
Moved example CRs to `examples/` dir

jewzaam:


lisa:
How do we feel about removing the examples for other folks who might want to use this operator?

jewzaam:
FYI https://issues.redhat.com/browse/OSD-3123

PR #27, opened 2020-03-16
Add machine_helper.go to utils package

csheremeta:
to house functions that may be useful for interacting with machine objects

### DIFF
--- a/pkg/controller/utils/machine_helper.go
+++ b/pkg/controller/utils/machine_helper.go
@@ -1,0 +1,125 @@
+package utils
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	awsproviderapi "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsproviderconfig/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var log = logf.Log.WithName("machine_helper")
+
+// RemoveAWSLBFromMasterMachines removes a Load Balancer (with name elbName) from
+// the spec.providerSpec.value.loadBalancers list for each of the master machine
+// objects in a cluster
+func RemoveAWSLBFromMasterMachines(kclient client.Client, elbName string, masterNodes *machineapi.MachineList) error {
+	for _, machine := range masterNodes.Items {
+		providerSpecDecoded, err := getAWSDecodedProviderSpec(machine)
+		if err != nil {
+			log.Error(err, "Error retrieving decoded ProviderSpec for machine", "machine", machine.Name)
+			return err
+		}
+		lbList := providerSpecDecoded.LoadBalancers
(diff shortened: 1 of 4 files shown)


sam-nguyen7:
/lgtm
This will make it a lot easier to unit test as well

pkg/controller/utils/machine_helper.go
```diff
@@ -0,0 +1,109 @@
+package utils
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	"github.com/prometheus/common/log"
```
lisa: How come we're using this for logging and not `sigs.k8s.io/controller-runtime/pkg/log`?

csheremeta: This switched on me when I saved the file, and I'm not sure why

lisa: Spoooooooky!

pkg/controller/utils/machine_helper.go
```diff
@@ -0,0 +1,111 @@
...
+	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	awsproviderapi "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsproviderconfig/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var log = logf.Log.WithName("machine_helper")
+
+func RemoveLBFromMasterMachines(kclient client.Client, elbName string, masterNodes *machineapi.MachineList) error {
```
lisa: Could you add a comment for this function? It should look like:

```
// RemoveLBFromMasterMachines ....
```

Check out [this blog post](https://blog.golang.org/godoc-documenting-go-code) for more on how Go wants code documented.

pkg/controller/utils/machine_helper.go
```diff
@@ -0,0 +1,111 @@
...
+		err = updateLBList(kclient, lbList, newLBList, machine, providerSpecDecoded)
+		if err != nil {
+			log.Error(err, "Error updating LB list for machine", "machine", machine.Name)
+			return err
+		}
+	}
+	return nil
+}
+
+func AddLBToMasterMachines(kclient client.Client, elbName string, masterNodes *machineapi.MachineList) error {
```
lisa: Doc here, too.

pkg/controller/utils/machine_helper.go
```diff
@@ -0,0 +1,111 @@
...
+		err = updateLBList(kclient, lbList, newLBList, machine, providerSpecDecoded)
+		if err != nil {
+			log.Error(err, "Error updating LB list for machine", "machine", machine.Name)
+			return err
+		}
+	}
+	return nil
+}
+
+func getDecodedProviderSpec(machine machineapi.Machine) (*awsproviderapi.AWSMachineProviderConfig, error) {
```
lisa: Less important for non-exported functions/methods, but might still be useful.

Also, this is specific to AWS, so does it make sense to name the function to follow suit?

PR #32, opened 2020-03-18
Refactor ingresscontroller logic from main reconcile function

sam-nguyen7:
Currently there is a lot of logic in our publishingStrategy reconcile function that can be refactor out. This PR focuses on the ingresscontroller logic. Splits into two function, one to handle default ingresscontroller, and one to handle non-default ingresscontroller.


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
 
+	"github.com/go-logr/logr"
```
lisa: Why not use the already imported logger? (line 28)

sam-nguyen7: removed

PR #42, opened 2020-03-25
OSD-3127: Fix AWS admin api SecurityGroup Reconciling

lisa:
Correct the logic of the AWS security group reconciling method. The
previous logic assumed that each IP Permission object (from AWS) would
ever only have a single CIDR block. That is, there would be a 1-to-1
mapping of CIDR blocks and IP Permission objects. This is not true.

The actual way the (rest of the) code works is that only a single IP
Permission object is created and within that object are multiple CIDR
blocks defined (`IpRanges`). As a result, the reconciliation algorithm
needs to iterate all of the objects: each IP Permission object, and each
IP range defined therein.

Signed-off-by: Lisa Seelye <lseelye@redhat.com>

### DIFF
--- a/pkg/awsclient/ec2_helper.go
+++ b/pkg/awsclient/ec2_helper.go
@@ -40,33 +40,59 @@ func (c *AwsClient) EnsureCIDRAccess(loadBalancerName, securityGroupName, vpcID 
 	for _, cidrBlock := range cidrBlocks {
 		seenExpectedRules[cidrBlock] = false
 	}
-	// For each ingress rule for the security group,
-Outer:
-	for _, ingressRule := range securityGroup.IpPermissions {
-		// Only care about 6443/TCP -> 6443/TCP
-		if *ingressRule.FromPort != 6443 &&
-			*ingressRule.ToPort != 6443 &&
-			*ingressRule.IpProtocol != "tcp" {
-			continue
-		}
-		// Since each ingressRule can have 0 or more IP Ranges defined, iterate through
-		// them for matches
-		for _, ipRange := range ingressRule.IpRanges {
-			for _, cidrBlock := range cidrBlocks {
-				if *ipRange.CidrIp == cidrBlock {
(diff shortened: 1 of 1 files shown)


pkg/awsclient/ec2_helper.go
```diff
@@ -49,13 +49,15 @@ Outer:
...
-				// No need to continue on this ingressRule, because we seen it
-				continue Outer
+		// Since each ingressRule can have 0 or more IP Ranges defined, iterate through
+		// them for matches
+		for _, ipRange := range ingressRule.IpRanges {
+			for _, cidrBlock := range cidrBlocks {
+				if *ipRange.CidrIp == cidrBlock {
+					seenExpectedRules[cidrBlock] = true
+					// No need to continue on this ingressRule, because we seen it
+					continue Outer
```
sam-nguyen7: what is Outer here?

lisa: `Outer` refers to the label on line 44, which is a bit outside the scope of what shows up in the diff.

Semantically it means: "If we have seen a CIDR block [that we want present] in the list that AWS gave us, then we can stop looking for it, and begin the next CIDR block [we want present]"

sam-nguyen7: ah makes sense. I didn't see the label

pkg/awsclient/ec2_helper.go
```diff
@@ -111,9 +139,6 @@ func (c *AwsClient) removeIngressRulesFromSecurityGroup(securityGroup *ec2.Secur
 		return nil
 	}
 	i := &ec2.RevokeSecurityGroupIngressInput{
-		FromPort:      aws.Int64(6443),
-		ToPort:        aws.Int64(6443),
-		IpProtocol:    aws.String("tcp"),
```
lisa: This is required because one may have either these attributes _or_ the `IpPermissions` attribute.

lisa:
/hold cancel

testing completed

PR #46, opened 2020-04-02
OSD-3189: First tests for APIScheme controller

lisa:
work in progress


pkg/awsclient/client.go
```diff
@@ -113,6 +113,39 @@ type Client interface {
 	DescribeSubnets(*ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error)
 	// CreateTags to apply tags to EC2 resources
 	CreateTags(*ec2.CreateTagsInput) (*ec2.CreateTagsOutput, error)
+
+	// Helper extensions
```
lisa, line 117: Added so that the _helper.go files are considered part of the overall interface so the mocked client will have them as well.

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -145,11 +146,13 @@ func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Res
...
-		NameSpace:  config.OperatorNamespace,
-		AwsRegion:  region,
-	})
+	if awsClient == nil {
+		awsClient, err = awsclient.GetAWSClient(r.client, awsclient.NewAwsClientInput{
+			SecretName: config.AWSSecretName,
+			NameSpace:  config.OperatorNamespace,
+			AwsRegion:  region,
+		})
+	}
```
lisa, line 155: I don't know if this is the best way to go about allowing the tests to set the client (line 32). It seems like it could be better.

lisa:
/hold cancel

I always get those backwards.

pkg/controller/apischeme/apischeme_controller.go
sam-nguyen7: Can't think of a way to make this better but it is annoying that we have to do this

pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+
+type mocks struct {
+	fakeKubeClient client.Client
+	mockCtrl       *gomock.Controller
+	mockAws        *awsclient.MockClient
+	scheme         *runtime.Scheme
+}
+
+// setupTests will make sure each test has the same starting point
+func setupTests(t *testing.T, localObjs []runtime.Object) *mocks {
```
sam-nguyen7, line 41: this is a great way to refactor out all the schemes from each tests. I will borrow this idea on my next unit tests :)

pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+		for i := range subnets {
+			if subnets[i] != test.mockSubnetIDOutput0[i] {
+				t.Fatalf("Wrong subnet ID back for public. Got %s, expected %s", subnets[i], test.mockSubnetIDOutput0[i])
+			}
+		}
+	}
+}
+
+func TestClusterBaseDomain(t *testing.T) {
```
sam-nguyen7, line 268: since this is testing `utils.GetClusterBaseDomain()` does it make more sense for this to be tested in the utils package?

pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+	base, err := utils.GetClusterBaseDomain(mocks.fakeKubeClient)
+	if err != nil {
+		t.Fatalf("Could not get cluster base domain name: %v", err)
+	}
+	if base != "unit.test" {
+		t.Fatalf("Base domain mismatch. Expected %s, got %s", "unit.test", base)
+	}
+}
+
+func TestMasterInstanceIds(t *testing.T) {
```
sam-nguyen7, line 288: same comment as above re utils package

pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+	expected := fmt.Sprintf("%s.%s", aObj.Spec.ManagementAPIServerIngress.DNSName, domain)
+	match := false
+	for _, cert := range api.Spec.ServingCerts.NamedCertificates {
+		for _, name := range cert.Names {
+			if name == expected {
+				match = true
+			}
+		}
+	}
+	if !match {
```
sam-nguyen7, line 355: another test idea is if the admin api endpoint already exist?

pkg/controller/apischeme/apischeme_controller_test.go
lisa: Yep, it is in fact in those tests (PR forthcoming). I think once they're committed, I'll remove the test here (and the other utils-focused tests).

PR #48, opened 2020-04-06
use hive.openshift.io/v1

staebler:


cblecker:
/assign @lisa 
I think this should go in, right?

lisa:
@csheremeta FYI for the cronjob.

/lgtm
/approve

PR #45, opened 2020-03-31
Unit tests for ingresscontroller logic

sam-nguyen7:
- Unit tests for helper functions written to handle ingresscontroller logic flow.
- Fixed an issue in one of the helper functions.

PR #49, opened 2020-04-07
Add manual testing of ingresscontroller instructions

sam-nguyen7:
Due to a race condition and how we handle these ingresscontrollers, it is difficult to write unit tests for them. This is an addition to https://github.com/openshift/cloud-ingress-operator/pull/45

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ It is possible to add additional applicationIngresses, however at this time, OSD
 
 ### Manual testing of default and nondefault ingresscontroller
 
-Due to a race condition with the [cluster-ingress-operator] (https://github.com/openshift/cluster-ingress-operator) we test the logic flow of ingresscontroller manually. Once you are in a cluster, here are the steps to do so:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,44 @@ Note: the `namespace` attribute for `secretRef` is not currently used; certifica
...
 It is possible to add additional applicationIngresses, however at this time, OSD supports the default plus an additional.
 
+## Testing
+
+### Manual testing of default and nondefault ingresscontroller
+
+Due to a race condition with the [cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator) we test the logic flow of ingresscontroller manually. Once you are in a cluster, here are the steps to do so:
+
+- Pause syncset to the cluster [SOP](https://github.com/openshift/ops-sop/blob/master/v4/howto/pause-syncset.md) 
+- Check to see which ingresscontrollers are present on the cluster with `oc get ingresscontrollers -n openshift-ingress-operator`
```
lisa, line 64: What is the check for?

sam-nguyen7: before and after results? I can remove this if you think it is not necessary

lisa: No, it's not that. Just want to make sure we tell folks what to be looking for, or what to do with the information they see. So if it's a "check the initial state before the test" or "make sure the cluster is in _some state we expect_" we should let the reader know.

PR #50, opened 2020-04-08
OSD-3190: Add unit tests for utils

lisa:
This adds the "testutils" helper package which is in use for the added
tests. Because there was going to be duplication in the apischeme
controller tests, those were consolidated and updated to reflect the new
helper.

Notably, utils/conditions.go is not tested (yet), and publishingstrategy
controller tests aren't unified with the helper at this time.


pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+const masterMachineLabel string = "machine.openshift.io/cluster-api-machine-role"
+
+const DefaultRegionName string = "us-east-1"
+const DefaultAzName string = "us-east-1a"
+const DefaultAPIEndpoint string = "https://api.unit.test:6443"
+const DefaultClusterDomain string = "unit.test"
+
+// ClusterTokenId represents the part of identifiers which is varied by the installer, eg
+// clustername-clustertokenid, as in load balancer names: foo-12345-us-east-1a
+const ClusterTokenId string = "12345"
```
2uasimojo, line 32: Curious why this doesn't use the grouping syntax
```
const (
    ...
)
```
...though I see it's just a straight cut/paste from the original.

pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -5,185 +5,16 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/utils/pointer"
-	awsprovider "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsproviderconfig/v1beta1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 const masterMachineLabel string = "machine.openshift.io/cluster-api-machine-role"
```
2uasimojo, line 16: This appears to have been put into testutils (L23) -- why isn't it removed from here?

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+	s := scheme.Scheme
+	if err := configv1.AddToScheme(s); err != nil {
+		t.Fatalf("Couldn't add configv1 scheme: (%v)", err)
+	}
+	if err := machineapi.AddToScheme(s); err != nil {
+		t.Fatalf("Couldn't add machine scheme: (%v)", err)
+	}
+	if err := cloudingressv1alpha1.SchemeBuilder.AddToScheme(s); err != nil {
+		t.Fatalf("Couldn't add cloudingressv1alpha1 scheme: (%v)", err)
+	}
```
2uasimojo, line 53: This method looks like a straight lift-and-shift from [1] -- except that this block used to be first. It looks like that's not a significant change logically, right?

No change suggested, just trying to understand why it moved.

[1] https://github.com/openshift/cloud-ingress-operator/pull/50/files#diff-e422e4e0f311f6f2a6e15f011a8e1b18L40-L62

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+				// <clustername>-<id>-ext
+				Name: fmt.Sprintf("%s-%s-ext", clusterid, ClusterTokenId),
+				Type: awsprovider.NetworkLoadBalancerType,
+			},
+			{
+				// <clustername>-<id>-int
+				Name: fmt.Sprintf("%s-%s-int", clusterid, ClusterTokenId),
+				Type: awsprovider.NetworkLoadBalancerType,
+			},
+		},
```
2uasimojo, line 128: nts: this is new, used by ValidateMachineLB (also added with this commit).

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+				DNSName:           dnsname,
+				AllowedCIDRBlocks: cidrs,
+			},
+		},
+	}
+}
+
+// ValidateMachineLB returns length, names and types (slices) and any error if one
+// The purpose is to have an easy way to condense 12+ lines of code
+func ValidateMachineLB(m *machineapi.Machine) (int, []string, []awsprovider.AWSLoadBalancerType, error) {
```
2uasimojo, line 208: nts: new

pkg/controller/utils/clusterinfo.go
```diff
@@ -61,6 +61,9 @@ func GetMasterNodeSubnets(kclient client.Client) (map[string]string, error) {
 	if err != nil {
 		return subnets, err
 	}
+	if len(machineList.Items) == 0 {
+		return subnets, fmt.Errorf("Did not find any master Machine objects")
+	}
```
2uasimojo, line 66: nts: this appears to be the only non-test change -- I guess a bug uncovered by the new tests.

pkg/controller/utils/clusterinfo_test.go
```diff
@@ -0,0 +1,201 @@
...
+	if v, ok := tags[expected]; ok {
+		if v != "owned" {
+			t.Fatalf("Expected owner tag %s to have value %s. Got %s", expected, "owned", v)
+		}
+	} else {
+		t.Fatalf("Expected to have a tag with key %s, but there wasn't one", expected)
+	}
+}
+
+// None of these should ever occur, but if they did, it'd be nice to know they return an error
```
2uasimojo, line 166: ++

pkg/controller/utils/clusterinfo_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	_, err = GetClusterPlatform(mocks.FakeKubeClient)
+	if err == nil {
+		t.Fatalf("Expected to get an error from not having an Infrastructure object")
+	}
+	_, err = GetClusterBaseDomain(mocks.FakeKubeClient)
+	if err == nil {
+		t.Fatalf("Expected to get an error from not having an Infrastructure object")
+	}
+
```
2uasimojo, line 200: Missing `GetMasterMachines`

2uasimojo:
Here are the "important" issues:
- https://github.com/openshift/cloud-ingress-operator/pull/50#discussion_r406509980
- https://github.com/openshift/cloud-ingress-operator/pull/50#discussion_r406522205
- https://github.com/openshift/cloud-ingress-operator/pull/50#discussion_r406452806

pkg/testutils/testutils.go
lisa: If I'm honest, it started out as one constant, then another popped up, and so I added it as another `const` line and didn't think about the grouping.

I'm happy to change the formatting, especially if it's better Go practice - I'm not entirely deep into the Go culture (yet).

pkg/controller/apischeme/apischeme_controller_test.go
lisa: Good catch. I'll remove once the Easter break is over.

pkg/testutils/testutils.go
lisa: The order changed for a rather mundane reason: The `testutils` package came about while writing for `utils` tests, and they didn't need the cloud ingress operator schemata. Once the `testutils` package was ready to be backported into the apischeme controller, I added it after the other scheme adds. No real importance.

pkg/controller/utils/clusterinfo.go
lisa: Yes, in one of the test cases I forgot to add `Machine` objects (a case which should never occur in OSD) and so I thought it best to handle in the `utils` package. When there are no `Machine` objects returned by the List method, there is a subscript reference as a few lines down the first object is referenced by `machineList.Items[0]`.

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+
+type Mocks struct {
+	FakeKubeClient client.Client
+	MockCtrl       *gomock.Controller
+	MockAws        *awsclient.MockClient
+	Scheme         *runtime.Scheme
+}
+
+// NewMockTest sets up for a new mock test, pass in some localObjs to seed the fake Kubernetes environment
+func NewTestMock(t *testing.T, localObjs []runtime.Object) *Mocks {
```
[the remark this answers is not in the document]
lisa: The struct is probably poorly named in that regard. `*Mocks` is in fact a singular object, and not, say `[]*Mock` (or even `*[]Mock`). The pluralization refers to the multiple mocked items inside the struct. I hope it isn't too confusing?

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+	awsconfig := &awsprovider.AWSMachineProviderConfig{}
+	err = codec.DecodeProviderSpec(&m.Spec.ProviderSpec, awsconfig)
+	if err != nil {
+		return l, names, lbTypes, err
+	}
+	for _, lb := range awsconfig.LoadBalancers {
+		names = append(names, lb.Name)
+		lbTypes = append(lbTypes, lb.Type)
+		l++
+	}
```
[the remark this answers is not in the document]
lisa: The reasoning behind the approach (which is entirely up for discussion) is so that in each case, the length being returned is always `l`, and not a fixed `0`. I have no idea if that is idiomatic or not, but it felt vaguely right. 🤷‍♀️

pkg/controller/utils/clusterinfo_test.go
```diff
@@ -0,0 +1,201 @@
...
+	mocks := testutils.NewTestMock(t, objs)
+	subnetmap, err := GetMasterNodeSubnets(mocks.FakeKubeClient)
+
+	if err != nil {
+		t.Fatalf("Couldn't get subnet map: %v", err)
+	}
+	if len(subnetmap) != 2 {
+		t.Fatalf("Expected Subnet map to have exactly two keys, but got %d keys back", len(subnetmap))
+	}
+	if _, ok := subnetmap["public"]; !ok {
```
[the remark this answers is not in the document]
lisa: The `_` (or whatever value there) is scoped to the `if` block. So to reuse later, something like this would be needed:

```go
var publicSubnet string
if publicSubnet, ok = subnetmap["public"]; !ok {
 t.Fatalf("failure")
}

if publicSubnet != expected {
  t.Fatalf("failure")
}
```

My opinion here is that semantically it isn't an improvement because the `publicSubnet` value is not useful anywhere; referring to it as `subnetmap["public"]` is fine.

pkg/testutils/testutils.go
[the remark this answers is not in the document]
2uasimojo: Yeah, this is a quirk of go that will take me some getting used to: that functions are set up to return both the values and the error. In cases where the error is set, you're supposed to ignore the other values because they're meaningless -- so it really wouldn't matter *what* you put in there. (I suppose there are (hopefully rare?) cases where the error is set, but the values are still meaningful, which is why the language authors set it up this way?)

So yeah, I don't know whether idiomatically it's frowned upon to do it the second way. A logically equivalent approach that perhaps straddles both sides of the fence would be to declare all the variables up front, but not initialize them; so they would get their so-called "zero value" -- which would be the same as the hardcoded values in my suggestion. Could even do the declaration as part of the signature and use naked `return` statements, which is even less hardcodey.

Anyway, like I said at the outset, definitely no change needed here, but I'm enjoying the discussion :)

pkg/controller/utils/machine_helper_test.go
```diff
@@ -0,0 +1,280 @@
...
+		t.Fatalf("Expected to have only 1 LoadBalancer, but got %d", len(awsconfig.LoadBalancers))
+	}
+	if awsconfig.LoadBalancers[0].Name != newLBList[0].Name {
+		t.Fatalf("Expected LB name to be %s, but got %s", newLBList[0].Name, awsconfig.LoadBalancers[0].Name)
+	}
+
+	if awsconfig.LoadBalancers[0].Type != newLBList[0].Type {
+		t.Fatalf("Expected LB type to be %s, but got %s", string(newLBList[0].Type), string(awsconfig.LoadBalancers[0].Type))
+	}
+}
```
csheremeta, line 85: we may also want a test here to ensure that no change is made if the two LB lists are the same

pkg/controller/utils/machine_helper_test.go
```diff
@@ -0,0 +1,280 @@
...
+			t.Fatalf("Can't lookup LB info: %v", err)
+		}
+
+		if l != 3 {
+			t.Fatalf("After the test we expect to have 3 load balancers, but got %d", l)
+		}
+		for _, lbType := range lbTypes {
+			if lbType != awsproviderapi.NetworkLoadBalancerType {
+				t.Fatalf("Expected to have a NLB, but got %s instead", string(lbType))
+			}
```
csheremeta, line 147: we may also want to look specifically for the LB named myelb here, with a found boolean

pkg/controller/utils/machine_helper_test.go
```diff
@@ -0,0 +1,280 @@
...
+				if lbName == test.nameToRemove {
+					found = true
+				}
+			}
+			if found {
+				t.Fatalf("Machine %s still has load balancer named %s", machineInfo.Name, test.nameToRemove)
+			}
+		}
+	}
+}
```
csheremeta, line 258: @2uasimojo has submitted a PR (awaiting the results of live testing) to change the RemoveAWSLBFromMasterMachines function to check for a match of the LB short name, instead of checking for the lblist to contain an LB with the same prefix as the short name we've found in the machine's provider spec (https://github.com/openshift/cloud-ingress-operator/pull/51)

originally I wrote the function to use strings.HasPrefix (line 30) because the short name would be something like "cluster-name-hash-ext" and the lb.Name would be "cluster-name-hash-ext-anotherhash-domain.example.com". however, if Eric's PR stands up to a live test (which is on my to-do list), we would probably want to write a test case where a machine has two LBs, one of which is named something that is the prefix of the other, in order to make sure that only the first gets removed, as expected.

for example, we'd name the LBs testlb-12345-ext and testlb-12345-ext-to-keep, and then attempt to remove just testlb-12345-ext. this sort of case is likely not to happen in the real world, which is why strings.HasPrefix has worked so far, but Eric's solution is better (assuming it passes live testing), as the current implementation would remove both LBs in this case.

pkg/controller/utils/machine_helper_test.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	newLBList := []awsproviderapi.LoadBalancerReference{
+		{
+			// Just something random!
+			Name: fmt.Sprintf("%s-%s-test", clustername, "12345"),
+			Type: awsproviderapi.NetworkLoadBalancerType,
+		},
+	}
+	// quickly make sure the test is going to measure an actual change
+	if len(oldLBList) != 2 || len(newLBList) != 1 {
```
sam-nguyen7, line 40: not sure this is needed?

lisa: The argument against is "Well, don't change the lists" and I hear that. The purpose of the assertion here is to ensure that the tested code is capable of going from two to one. But say the `oldLBList` is of length 1, or even 0, the test is going to behave differently.

I guess that's a roundabout way of saying the test should be more flexible with multiple test cases to _validate_ the different behaviour, eh?

pkg/controller/utils/clusterinfo_test.go
```diff
@@ -0,0 +1,201 @@
...
+	objs := []runtime.Object{infraObj, machineList}
+	mocks := testutils.NewTestMock(t, objs)
+
+	_, err := GetMasterNodeSubnets(mocks.FakeKubeClient)
+	if err == nil {
+		t.Fatalf("Expected to see an exception when trying to get subnets for 0 master nodes ")
+	}
+}
+
+func TestGetClusterRegion(t *testing.T) {
```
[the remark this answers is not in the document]
lisa: Good question! According to the code it should return an `error`, but it isn't tested here. I can add a case for that?

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,227 @@
...
+			Kind:       "AWSMachineProviderConfig",
+		},
+		InstanceType:       "small",
+		BlockDevices:       []awsprovider.BlockDeviceMappingSpec{},
+		AMI:                awsprovider.AWSResourceReference{ID: &ami},
+		Tags:               []awsprovider.TagSpecification{{Name: fmt.Sprintf("kubernetes.io/cluster/%s", clusterid), Value: "owned"}},
+		IAMInstanceProfile: &awsprovider.AWSResourceReference{ID: pointer.StringPtr(fmt.Sprintf("%s-%s-profile", clusterid, role))},
+		UserDataSecret:     &corev1.LocalObjectReference{Name: "aws-cloud-credentials"},
+		Placement:          awsprovider.Placement{Region: region, AvailabilityZone: zone},
+		LoadBalancers: []awsprovider.LoadBalancerReference{
```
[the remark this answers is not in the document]
lisa: That is weird, but a your change is reverted with `gofmt`. I'm not quite sure what's going on in the `gofmt` internals 🤷‍♀️ , so I'll reject this suggestion since I don't want to fight `gofmt`.

pkg/testutils/testutils.go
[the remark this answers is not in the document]
lisa: I disagree on that one. This function is only meant to do the tally, not to determine if the resulting numbers are right or wrong.

pkg/testutils/testutils.go
lisa: Then I am happy to leave it for Future People to change if they so desire.

pkg/controller/utils/machine_helper_test.go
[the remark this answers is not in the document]
lisa: > we may also want a test here to ensure that no change is made if the two LB lists are the same

Wouldn't this test still handle it? If `newLBList` is treated as the authoritative set of load balancers for the machines in question we would expect the code to always ensure that the machine load balancers are equal to that set. If it happens that `newLBList` and oldLBList` are the same, isn't the `updateAWSLBList` code effectively idempotent because at the end of the worst case, the machine load balancers will be removed and re-added with the same values?

lisa: > we may also want a test here to ensure that no change is made if the two LB lists are the same

Wouldn't this test still handle it? If `newLBList` is treated as the authoritative set of load balancers for the machines in question we would expect the code to always ensure that the machine load balancers are equal to that set. If it happens that `newLBList` and oldLBList` are the same, isn't the `updateAWSLBList` code effectively idempotent because at the end of the worst case, the machine load balancers will be removed and re-added with the same values?

pkg/controller/utils/machine_helper_test.go
```diff
@@ -0,0 +1,280 @@
...
+	machineInfo := types.NamespacedName{
+		Name:      sampleMachine.GetName(),
+		Namespace: sampleMachine.GetNamespace(),
+	}
+	// reload the object to make sure we're not just working with the "in-memory"
+	// representation, that being, the un-saved one.
+	err = mocks.FakeKubeClient.Get(context.TODO(), machineInfo, &sampleMachine)
+	if err != nil {
+		t.Fatalf("Couldn't reload the test machine: %v", err)
+	}
```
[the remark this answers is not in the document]
lisa: So it "go"es :(

pkg/controller/utils/clusterinfo_test.go
lisa: `GetMasterMachines` does not rely on an `Infrastructure` object, so it doesn't need to be included here.

pkg/testutils/testutils.go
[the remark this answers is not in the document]
2uasimojo: Odd indeed.

pkg/testutils/testutils.go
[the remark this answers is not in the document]
2uasimojo: `Validate` seems an odd choice of names in that case.

I agree it's not determining if the resulting numbers are right or wrong, but it still has error paths. As written, the caller is (always) triggering test failure in response to those errors; I'm just suggesting moving that logic into the helper.

But meh.

lisa:
This is going to be a github squash merge, so please don't remove the hold.

2uasimojo:
I'd still like to see https://github.com/openshift/cloud-ingress-operator/pull/50#discussion_r406509980 fixed before we merge this, since it's an actual logic error.

pkg/controller/utils/machine_helper_test.go
```diff
@@ -0,0 +1,280 @@
...
+			}
+		}
+
+		// Make change
+		err := RemoveAWSLBFromMasterMachines(mocks.FakeKubeClient, test.nameToRemove, machineList)
+		if err != nil {
+			if !test.shouldFail {
+				t.Fatalf("Unexpected test couldn't remove LB %s from Machine: %v", test.nameToRemove, err)
+			}
+		}
```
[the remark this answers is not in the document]
lisa: @2uasimojo I see two suggestions here. Can you edit your comment to reflect which you'd like to make?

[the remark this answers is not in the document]
2uasimojo: As noted, the first suggestion breaks the test, which is what exposed the bug. So the second suggestion is the correct one (we don't expect the call to error in either case).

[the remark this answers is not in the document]
lisa: @2uasimojo Does 6866d3f cover what you were after?

2uasimojo:
Thanks @lisa.
/lgtm but I'll leave the hold until @csheremeta is happy.

pkg/controller/utils/machine_helper_test.go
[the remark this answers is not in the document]
2uasimojo: Yup, thanks.

pkg/controller/utils/machine_helper_test.go
csheremeta: this makes good sense to me

2uasimojo: I may be misunderstanding, but it sounds like we're accepting the missing test based on knowing how the underlying code works. IOW: Yes, it's idempotent as written, but the tests shouldn't know that or rely on it. The point of these tests is to validate the visible behavior. If the code changes its algorithm later -- say, to do piecemeal add/remove rather than wholesale replace -- then an additional test like the one @csheremeta suggests are important to make sure that algorithm is correct.

lisa: No, what I'm saying is that the behaviour of the method being tested is that the `newLBList` is always the authoritative source of what the Machine in question should have as its load balancer reality. Because of this, it is nonsensical to test what happens when old == new because at the end of the method, everything at all times will be set to the new list, and a test with a difference between old and new will cover just the same because of the end state this method promises: all will become equal to the new list's contents.

pkg/controller/utils/machine_helper_test.go
lisa: Maybe we can address that in a rebased #51?

pkg/controller/utils/machine_helper_test.go
2uasimojo: We're agreeing that the method under test is always replacing with `newLBList`. If we're just interested in white-box/unit testing, what you're saying is legit.

But the point of writing tests we know are going to pass (they have to pass before the PR merges) is to catch bugs when the underlying code is changed. And to that end, there's value in writing the suggested test case, because it does cover a different _black-box_/interface behavior.

I'm not asking for a change here (or necessarily at all). I'm just asserting that it's not worthless.

PR #51, opened 2020-04-09
Match LB short name for removal

2uasimojo:
Previously when removing a load balancer, we were matching a substring
of its DNS name. With this commit, we pass in the "short name" so we can
do an exact match.


2uasimojo:
/hold

This needs live test.

csheremeta:
I finally got around to live testing this today. Sorry for the delay. We'll want to come back to the unit test noted in https://github.com/openshift/cloud-ingress-operator/pull/50#discussion_r407670120
/lgtm

2uasimojo:
I'll get to those tests in a subsequent PR. Thanks for testing. Let's land this.
/hold cancel

2uasimojo:
/approve
?

PR #52, opened 2020-04-09
Add comments for {Add|Remove}AWSLB{To|From}MasterMachines

2uasimojo:
machine_helper.AddAWSLBToMasterMachines and
.RemoveAWSLBFromMasterMachines both take an argument called `elbName`,
but are expecting different versions of the LB name for $reasons. This
was confusing to me, reading the code for the first time, so this commit
adds explanatory comments so future me isn't as confused.


2uasimojo:
#51 works, so this isn't needed \o/

PR #53, opened 2020-04-13
Add tags for external NLB

sam-nguyen7:
Currently when creating an external NLB we do not have any tags. This PR fixes that issue by creating two appropriate tags at creation time of the external NLB


sam-nguyen7:
/hold 
for testing

sam-nguyen7:
/hold cancel
this needs to be tested on stage

PR #54, opened 2020-04-16
OSD-3391, BZ 1814332: Handle missing infra PlatformStatus

lisa:
Some older, upgraded clusters currently lack the `PlatformStatus` in the
`Infrastructure` object, and so this works around it by referencing the
(deprecated) cluster-config-v1 configmap out of kube-system namespace.

### DIFF
--- a/deploy/20_cluster_config_v1_reader_role.yaml
+++ b/deploy/20_cluster_config_v1_reader_role.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: cluster-config-v1-reader
+  namespace: kube-system
+rules:
+- apiGroups:
+  - ""
+  resourceNames:
+  - cluster-config-v1
+  resources:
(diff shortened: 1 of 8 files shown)


deploy/20_cluster_config_v1_reader_role.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: cluster-config-v1-reader
```
cblecker, line 5: As the managed-velero-operator creates this exact named role, will this conflict?

lisa: I was thinking about that same thing, and I'm not sure. Both operators need to ensure that the role is there, but can both try to create it in the same way?

lisa: I actually think I need to also update the selectorsyncset here.

lisa:
/retest

uh oh?

PR #56, opened 2020-04-27
Support deletion of non-default ingresscontroller and fix publishingStrategy logs

sam-nguyen7:
- Add annotations to all ingresscontrollers created by our operator. This enables us to delete non-default ingresscontroller when publishing CR is changed/edited.
- Fix example CR to better reflect state of publishingStrategy.
- Change logs to reduce noise from operator pod.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -116,6 +116,23 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	// get a list of all ingress on cluster that has annotation owner cloud-ingress-operator
+	// and delete all non-default ingresses
+	for _, ingressController := range ingressControllerList.Items {
+		if ingressController.Name != "default" && ingressController.Annotations["Owner"] == "cloud-ingress-operator" {
+			log.Info(fmt.Sprintf("ingresscontroller to be deleted: %v", ingressController))
+			err := r.client.Delete(context.TODO(), &ingressController)
+			if err != nil {
+				log.Error(err, "failed to delete ingresscontroller")
+				return reconcile.Result{}, err
+			}
+		}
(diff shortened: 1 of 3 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -116,6 +116,23 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+			if err != nil {
+				log.Error(err, "failed to delete ingresscontroller")
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	// wait 60 seconds for deletion to be completed
+	log.Info("waited 60 seconds for necessary ingresscontroller deletions")
+	time.Sleep(time.Duration(60) * time.Second)
```
lisa, line 134: Is this a good candidate for `return reconcile.Result{RequeueAfter: time.Duration(60 * time.Second)}, nil` so the operator can handle other things?

sam-nguyen7: +1. However, we need to have the proper status set up. I will write up a card and come back to this once we have status and conditions for the publishingStrategy CR

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -404,30 +416,10 @@ func (r *ReconcilePublishingStrategy) nonDefaultIngressHandle(appingress cloudin
...
-			}
-		} else {
-			log.Error(err, fmt.Sprintf("failed to create new ingresscontroller with domain %s", appingress.DNSName))
-			return err
-		}
+		log.Error(err, fmt.Sprintf("got error trying to create %s", newIngressController.GetName()))
+		return err
 	}
-	log.Info("successfully created new ingresscontroller")
+	//log.Info("successfully created new non-default ingresscontroller")
```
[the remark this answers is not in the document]
sam-nguyen7: :+1:

PR #59, opened 2020-05-14
This remove adding rh-api endpoint to APIServer cluster config logic

wanghaoran1988:
hive already added.


lisa:
Good find @wanghaoran1988 and @gsleeman. cc @igoihman 

/approve
/lgtm

lisa:
For future reference, Hive adds this here: https://github.com/openshift/hive/blob/925f9d3f8dc0056fbdf070e76fc3260f472581db/pkg/controller/controlplanecerts/controlplanecerts_controller.go#L314

PR #62, opened 2020-05-14
Fix deletion for ingress

sam-nguyen7:
Currently deletion of ingresscontroller are being triggered every reconcile loop. This PR fixes that by:

- only deleting if necessary (e.g ingress listening has been changed).
- only delete non-default ingress with correct annotation.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -116,7 +116,8 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	err = r.deleteIngressWithAnnotation(ingressControllerList)
+	// delete non-default with proper annotation if not publishingStratagy CR
+	err = r.deleteIngressWithAnnotation(instance.Spec.ApplicationIngress, ingressControllerList)
 	if err != nil {
 		log.Error(err, "Cannot delete ingresscontroller with annotation")
 		return reconcile.Result{}, err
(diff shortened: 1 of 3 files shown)


sam-nguyen7:
/hold 
fixing unit tests

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -351,18 +352,38 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		}
+	}
+
+	for _, ingress := range toBeDeleted {
+		err := r.client.Delete(context.TODO(), ingress)
+		if err != nil {
+			log.Error(err, "Failed to delete ingresscontroller")
+			return err
+		}
+	}
```
lisa, line 370: Can this be consolidated to do the deletion in the first for loop?

pkg/controller/publishingstrategy/publishingstrategy_controller_test.go
```diff
@@ -378,21 +396,75 @@ func TestDeleteIngressWithAnnotation(t *testing.T) {
...
+	// if ingress without annotation hit method, then it should not be removed
+	err = r.deleteIngressWithAnnotation(mockPublishingStrategy().Spec.ApplicationIngress, ingressControllerList)
+	if err != nil {
+		t.Fatalf("couldn't delete ingress")
+	}
+
+	if ingressControllerList.Items[0].Spec.Domain != "apps2.exaple-nondefault-domain-with-no-annotation" {
+		t.Fatalf("expect nondefault ingress to be on cluster but it is not")
+	}
+}
```
lisa, line 440: Will `ingressControllerList` be updated by the `Delete` in the `deleteIngressWithAnnotation` method, or do you need to do an explicit `Get` from the (fake) Kubernetes client to check for the no annotation exists?

PR #63, opened 2020-05-19
OSD-3564: Use native Service object to create ELB

lisa:
work in progress to use native k8s Service objects to create the rh-api ELB.


lisa:
No, bot, really, this is a work in progress.

fahlmant:
If you put [WIP] at the start of the title, the bot will add it

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -61,6 +62,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
+	// Watch the Service we set up for rh-api
+	err = c.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &cloudingressv1alpha1.APIScheme{},
+	})
+	if err != nil {
+		return err
+	}
+
```
lisa, line 73: Presumably because there are no owner references on that `Service`, this can go away.

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -196,132 +184,78 @@ func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Res
...
 		r.client.Status().Update(context.TODO(), instance)
 		return reconcile.Result{}, err
+
 	}
 
 	SetAPISchemeStatus(instance, "Success", "Admin API Endpoint created", cloudingressv1alpha1.ConditionReady)
 	r.client.Status().Update(context.TODO(), instance)
-
-	return reconcile.Result{}, nil
+	reqLogger.Info("Reconciling in 60 seconds...")
```
lisa, line 224:
```suggestion
```

No need to fill the logs?

cblecker:
/uncc
/cc @luis-falcon @mbarnes 

Thanks for ccing me on this one, Lisa.
Luis, Matt: This logic will apply to the rh-ssh endpoint too.

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -127,12 +112,39 @@ func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Res
...
+			// Reconcile again to get the new Service and give AWS time to create the ELB
+			reqLogger.Info("Service was just created, so let's try to requeue to set it up")
+			return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		} else if err != nil {
+			reqLogger.Error(err, "Couldn't get the Service")
+			return reconcile.Result{}, err
+		}
+	}
+	// Reconcile the access list in the Service
+	if !sliceEquals(found.Spec.LoadBalancerSourceRanges, instance.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks) {
```
mbarnes, line 136: Just had a question about this part, coming from how we do things in `managed-velero-operator`.
There, the pattern is usually to create the `dep` object just before `r.client.Get()`, and then at this point do a deep equal between `dep` and `found`.

Was avoiding a full `Service` comparison intentional?

`dep.Spec.LoadBalancerSourceRanges` is the same as `instance.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks`, so you'd get the same CIDR mismatch detection by comparing `dep` and `found`, but also catch any other inconsistencies in `found`.

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -127,12 +112,39 @@ func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Res
...
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Spec.ManagementAPIServerIngress.DNSName, Namespace: "openshift-kube-apiserver"}, found)
 	if err != nil {
-		reqLogger.Error(err, "Couldn't get the cluster owner tags")
-		SetAPISchemeStatus(instance, "Couldn't reconcile", "Couldn't get the cluster owner tags", cloudingressv1alpha1.ConditionError)
-		r.client.Status().Update(context.TODO(), instance)
-		return reconcile.Result{}, err
+		if errors.IsNotFound(err) {
+			// need to create it
+			dep := r.newServiceFor(instance)
+			err = r.client.Create(context.TODO(), dep)
```
mbarnes, line 122: Does the new object need an owner reference?

lisa: @mbarnes Strictly speaking it would be nice to have, but [cross-namespace ownership doesn't work](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents):

> *Note*: Cross-namespace owner references are disallowed by design. This means: 1) Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped. 2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.

pkg/controller/apischeme/apischeme_controller.go
mbarnes: @lisa answered privately:
> it was intentional because the thing we care about is only the allowed cidr blocks, which is a []string. returning a new service object only to compare that seems excessive.

PR #67, opened 2020-05-28
Fix contains logic

sam-nguyen7:
Fix contains function to better match the ingress that should be deleted. 
Fix bug of deleting default ingress.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -358,6 +358,13 @@ func (r *ReconcilePublishingStrategy) deleteIngressWithAnnotation(appIngressList
 		if _, ok := ingress.Annotations["Owner"]; !ok {
 			continue
 		}
+		// if ingress is default, skip it since we are only looking at non-default ingresses
+		if ingress.Name == "default" {
+			continue
+		}
+		// only delete the ingress that does not exist on publishingStrategy
+		// true means the ingress (on cluster) exists on the publishingStrategy CR as well, so no deletion
+		// false means the ingress (on cluster) do not exist on the CR, so delete to have desired result
 		if !contains(appIngressList, &ingress) {
 			err := r.client.Delete(context.TODO(), &ingress)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -358,6 +358,13 @@ func (r *ReconcilePublishingStrategy) deleteIngressWithAnnotation(appIngressList
 		if _, ok := ingress.Annotations["Owner"]; !ok {
 			continue
 		}
+		// if ingress is default, continue as we are only looking at non-default ingresses
```
lisa, line 361:
```suggestion
		// if ingress is default, skip it since we are only looking at non-default ingresses
```

`continue` is a Go keyword, but also could mean "continue with the flow of code" as in, something desired. So minor language suggestion here.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -358,6 +358,13 @@ func (r *ReconcilePublishingStrategy) deleteIngressWithAnnotation(appIngressList
...
+		if ingress.Name == "default" {
+			continue
+		}
+		// only delete the ingress that does not exist on publishingStrategy
+		// true means the ingress (on cluster) exists on the publishingStrategy CR as well, so no deletion
+		// false means the ingress (on cluster) do not exist on the CR, so delete to have desired result
```
lisa, line 367: Doesn't the `contains` function also check for changes?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -370,18 +377,29 @@ func (r *ReconcilePublishingStrategy) deleteIngressWithAnnotation(appIngressList
...
+// this helps to determine if ingress needs to be deleted or not. The end goal is to have all ApplicationIngress on PublishingStrategy CR on cluster
 func contains(appIngressList []cloudingressv1alpha1.ApplicationIngress, ingressController *operatorv1.IngressController) bool {
+	var isContained bool
 	for _, app := range appIngressList {
+		log.Info(fmt.Sprintf("app being processed %s", app.DNSName))
+		// TODO (sam): Add a comment about why this if is here, and the above method too
+		// if the ApplicationIngress (on CR) is default then set bool to false
+		// eg. ingresscontroller (on cluster): apps2
+		//     appIngressList (in CR) : [default]
+		// since apps2 does NOT exist in appIngressList, set bool to false and ready for deletion
```
lisa, line 391:
```suggestion
		// if the ApplicationIngress (on CR) is default then set bool to false
		// eg. ingresscontroller (on cluster): apps2
		//     appIngressList (in CR) : [default]
		// since apps2 does NOT exist in appIngressList, set bool to false and ready for deletion
```

Looks like you wrote your comment? ;)

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: no it does not. Changes to the specific ingress is handled elsewhere in the controller.

lisa: Line 395-399 (new revision) has the check for dns name changes though?

sam-nguyen7: that only checks if the ingress.dnsName (on cluster) matches with any app.dnsName (in CR). (eg if "apps2.non-default-domain" (on cluster) actually matches with "apps2.non-default-domain" (in CR)). 

So if apps2 on cluster is set to "internal" while apps2 in CR is set to "external, specific changes like this are NOT handled here, as we deal with that later on.

PR #69, opened 2020-06-02
Fix RouteSelector check

sam-nguyen7:
- Fix RouteSelector check which causes non-default ingress to be deleted
- Add error check fro creation of non-default ingress

Addresses OHSS-62

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -15,7 +15,7 @@ parameters:
 metadata:
   name: selectorsyncset-template
 objects:
-- apiVersion: hive.openshift.io/v1alpha1
+- apiVersion: hive.openshift.io/v1
   kind: SelectorSyncSet
   metadata:
     labels:
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -466,8 +466,8 @@ func (r *ReconcilePublishingStrategy) nonDefaultIngressHandle(appingress cloudin
 	err = r.client.Create(context.TODO(), newIngressController)
 	if err != nil {
 		if k8serr.IsAlreadyExists(err) {
-			for i := 1; i < 60; i++ {
-				if i == 60 {
+			for i := 1; i < 24; i++ {
+				if i == 24 {
 					log.Error(err, "out of retries to create non-default ingress")
 				}
 				time.Sleep(time.Duration(i) * time.Second)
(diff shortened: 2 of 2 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -15,7 +15,7 @@ parameters:
 metadata:
   name: selectorsyncset-template
 objects:
-- apiVersion: hive.openshift.io/v1alpha1
```
fahlmant, line 18: What's the reason for this change?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -273,7 +273,7 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 			log.Error(err, "cannot get infrastructure name")
 			return reconcile.Result{}, err
 		}
-		extNLBName := infrastructureName + "-test"
+		extNLBName := infrastructureName + "-ext"
```
fahlmant, line 276: Reason for change? Does this affect the logic changing below?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -545,12 +559,14 @@ func isOnCluster(publishingStrategyIngress *cloudingressv1alpha1.ApplicationIngr
 		return false
 	}
 	if publishingStrategyIngress.RouteSelector.MatchLabels != nil {
```
fahlmant, line 561: Should this be an or instead of nested?

hack/olm-registry/olm-artifacts-template.yaml
cblecker: We should be using hive/v1 everywhere by now

fahlmant: Has the operator been tested with hive v1?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: this was found by Will Gordon. Very small change. Does not change logic below.

hack/olm-registry/olm-artifacts-template.yaml
sam-nguyen7: https://github.com/openshift/cloud-ingress-operator/pull/48 this PR has been merged awhile back. It was just never generated. afaik it's OK be be rolled out

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: Should this be done in a seperate PR or commit then?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
lisa: I agree with @fahlmant on this one. Can this be re-worked to avoid the nested conditionals?

hack/olm-registry/olm-artifacts-template.yaml
fahlmant: IMO should be in a different commit/PR since if you revert this PR this will be reverted as well. But I won't block PR on this

lisa: I agree with Taylor here, I wouldn't 👎 but if we can easily split it out, we should.

lisa:
@sam-nguyen7 Before we promote this let's talk. We might need to revert #63 if the behaviour can't be quickly understood.

PR #72, opened 2020-06-04
See #63: Re-commit for testing purposes

lisa:
See #63 (23b1035) and #70.

This is functionally the same as #63, but it includes more logging statements to see what's going on.


fahlmant:
Quick question:
Why not test this PR on target clusters? If this is known to be semi-broken, why merge it so all stage clusters get it? This could interfere with other teams testing

lisa:
> Quick question:
> Why not test this PR on target clusters? If this is known to be semi-broken, why merge it so all stage clusters get it? This could interfere with other teams testing

Right now we know there's a problem with new clusters being created. It would be hard to replicate that if this is installed after a cluster is created. I don't know if it's possible to do that short of merging.

PR #73, opened 2020-06-05
OSD-3564: Remove extra log statements & Spec is mandatory

lisa:
This change is wider scoped than only APIScheme because it also makes
Spec mandatory for PublishingStrategy CRDs. We always want these to be
mandatory since we need them present to do anything useful in their
respective controllers.

In order to keep tabs on some of what is going on in the APIScheme
controller, there is a new variable to control when to print it's
activities. It will log activity every 1000 reconcilations (every ~1000
minutes), just so we have some other kind of heartbeat.

### DIFF
--- a/deploy/crds/cloudingress.managed.openshift.io_apischemes_crd.yaml
+++ b/deploy/crds/cloudingress.managed.openshift.io_apischemes_crd.yaml
@@ -19,12 +19,12 @@ spec:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
             of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
           type: string
         kind:
           description: 'Kind is a string value representing the REST resource this
             object represents. Servers may infer this from the endpoint the client
-            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
           type: string
         metadata:
           type: object
@@ -38,12 +38,17 @@ spec:
                 https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html'
               properties:
                 allowedCIDRBlocks:
+                  description: AllowedCIDRBlocks is the list of CIDR blocks that should
+                    be allowed to access the management API
                   items:
                     type: string
                   type: array
                 dnsName:
+                  description: DNSName is the name that should be used for DNS of
+                    the management API, eg rh-api
                   type: string
                 enabled:
+                  description: Enabled to create the Management API endpoint or not.
                   type: boolean
               required:
               - allowedCIDRBlocks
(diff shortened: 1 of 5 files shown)


lisa:
/cc @sam-nguyen7 

This changes `PublishingStrategy` CRDs. Thoughts?

deploy/crds/cloudingress.managed.openshift.io_apischemes_crd.yaml
```diff
@@ -62,7 +67,50 @@ spec:
                 code after modifying this file Add custom validation using kubebuilder
                 tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html'
               type: string
+            conditions:
```
sam-nguyen7, line 70: how are these generated? I don't see any modification in the _types.go file to reflect this generation. Was it in another PR?

deploy/crds/cloudingress.managed.openshift.io_publishingstrategies_crd.yaml
```diff
@@ -128,6 +128,8 @@ spec:
         status:
           description: PublishingStrategyStatus defines the observed state of PublishingStrategy
           type: object
+      required:
```
sam-nguyen7, line 131: lgtm

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -90,8 +96,10 @@ type LoadBalancer struct {
 // 3. Ready for work (Ready)
 func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-
-	reqLogger.Info("Reconciling APIScheme", "request", request)
+	reconcileCount++
+	if reconcileCount%1000 == 0 {
+		reqLogger.Info("Reconciling APIScheme", "request", request)
+	}
```
sam-nguyen7, line 102: what is the purpose of this?

fahlmant: I don't think this will carry over between reconciles.

deploy/crds/cloudingress.managed.openshift.io_apischemes_crd.yaml
lisa: Yes. This was a result of not running `operator-sdk generate ...` commands a long time ago after the status/conditions PR.

pkg/controller/apischeme/apischeme_controller.go
lisa: We don't want to log every reconcile because if we did that, it would fill the logs every minute, so this is a middle ground between zero log entries and too many.

This snippet will log a line for every 1000th reconcile, or approximately every 1000 minutes.

lisa: @fahlmant Won't it? Up on line 36 `reconcileCount` is defined as a package variable, its value should carry?

fahlmant: Ah I missed that bit.

Still, I think this is the wrong way to go. Once a minute is not that much logging. And taking out any logs for the rest of the reconcile reduces visibility into what the operator is doing to what resources.

lisa: @fahlmant I think you're right. It's probably better to err on the side of "too much" than "not enough" because we have the option to turn the knob down later. I'll make that change here.

/hold

lisa: /hold cancel

@fahlmant @sam-nguyen7 thoughts on c9e363e?

2uasimojo: Just a drive-by here...

Why is the `Reconcile` being called so frequently in the first place? Is something legitimately changing that's triggering it?

lisa: @2uasimojo Yes, on line 221 we intentionally requeue. The purpose is to ensure that any changes to the `openshift-kube-apiserver/service/rh-api` object's `LoadBalancerSourceRanges` are reconciled with the authoritative source (the apischeme CR); see lines 142-155.

lisa:
/hold

for github squash and merge

PR #76, opened 2020-06-30
Add controller for SSH pod

mbarnes:
This adds a CRD and controller for an SSH pod as demoed by @mwoodson to allow SRE access to private AWS clusters as themselves rather than as system admin.

Support for GCP to be added in the future.


mbarnes:
We'll need to mimic the `APIScheme` changes from [OSD-3209](https://issues.redhat.com/browse/OSD-3209).

mbarnes:
This is ready for review now, removing "do-not-merge" labels.

deploy/80_rh-ssh.Namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "${{SSHD_NAMESPACE}}"
```
lisa, line 4: Where does SSHD_NAMESPACE come from? What is the value expected to be?

deploy/90_rh-ssh.SSHD.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: cloudingress.managed.openshift.io/v1alpha1
+kind: SSHD
+metadata:
+  name: rh-ssh
+  namespace: "${{SSHD_NAMESPACE}}"
+spec:
+  dnsName: rh-ssh
+  allowedCIDRBlocks: "${{ALLOWED_CIDR_BLOCKS}}"
```
lisa, line 8: @csheremeta how does this line up with the work you've done around this piece?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/aws/aws-sdk-go v1.20.19
 	github.com/coreos/prometheus-operator v0.35.1
+	github.com/go-logr/logr v0.1.0
```
lisa, line 8: Please remove this and use the existing log facility.

pkg/apis/cloudingress/v1alpha1/sshd_types.go
```diff
@@ -0,0 +1,78 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
```
lisa, line 8: These comments can be removed.
```suggestion
```

pkg/apis/cloudingress/v1alpha1/sshd_types.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// ConfigMapSelector is a label selector to isolate config maps containing SSH authorized keys
+	// to be mounted into the SSHD container
+	ConfigMapSelector metav1.LabelSelector `json:"configMapSelector,omitempty"`
+}
+
+// SSHDStatus defines the observed state of SSHD
+// +k8s:openapi-gen=true
+type SSHDStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
```
lisa, line 43:
```suggestion
```

This comment can be removed.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
```
lisa, line 12: Please remove this and use the existing log facility imported as `logf`.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+		client.InNamespace(instance.Namespace),
+		&client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		r.SetSSHDStatusError(instance, reqLogger, "Failed to list config maps with SSH keys", err)
+		return reconcile.Result{}, err
+	}
+	// Sort ConfigMaps by name for the purpose of creating
+	// stable Volume and VolumeMount lists in the Deployment.
+	sort.Slice(configMapList.Items, func(i, j int) bool {
+		return configMapList.Items[i].Name < configMapList.Items[j].Name
+	})
```
lisa, line 197: This bit seems to be wanting a custom resource to represent a SSH key. I'd love to see that so we can talk about SSH keys and not config maps.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+					return reconcile.Result{Requeue: true}, nil
+				}
+				r.SetSSHDStatusError(instance, reqLogger, "Failed to create deployment", err)
+				return reconcile.Result{}, err
+			}
+		} else {
+			return reconcile.Result{}, err
+		}
+	} else {
+		// Deployment exists, check if it's updated.
```
lisa, line 225: What does "updated" mean in the context of this comment?

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+		if !reflect.DeepEqual(foundDeployment.Spec, deployment.Spec) {
+			// Specs aren't equal, update and fix.
```
lisa, line 227: What does "equal" mean here? Why should they be? In other words, I can see that the line above is comparing the equality (or inequality, to be more precise) but why is that being done?

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+				return reconcile.Result{}, err
+			}
+			// Requeue to give AWS time to apply the changes.
+			reqLogger.Info("Requeuing after service update")
+			return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		}
+	}
+
+	// Create a Route53 DNS entry for the service's load balancer.
+	// FIXME Consider using https://github.com/openshift/external-dns
```
lisa, line 287:
```suggestion
	// TODO Consider using https://github.com/openshift/external-dns
```

Minor nit - Not really a FIXME, more of a to do item, and 👍  for using it when it comes around.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+
+func getMatchLabels(cr *cloudingressv1alpha1.SSHD) map[string]string {
+	return map[string]string{"deployment": cr.Name}
+}
+
+func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.ConfigMapList) *appsv1.Deployment {
+	replicas := int32(1)
+	terminationGracePeriod := int64(30) // seconds
+	revisionHistoryLimit := int32(10)
+	progressDeadline := int32(600) // seconds
```
lisa, line 342: Please consider using `"k8s.io/utils/pointer"` and then in-place `pointer.Int32ptr(1)`, pointer.Int64ptr(30)` and so on.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+	return &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Deployment",
+			APIVersion: appsv1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cr.Name,
+			Namespace: cr.Namespace,
+		},
+		Spec: appsv1.DeploymentSpec{
```
lisa, line 394: I think we will want to restrict the kinds of nodes on which this runs to at least infra and probably only on master nodes. Please add the necessary tolerations/nodeAffinity.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	// Watch for changes to Services
+	err = c.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &cloudingressv1alpha1.SSHD{},
+	})
+	if err != nil {
+		return err
+	}
```
lisa, line 80: One caution is that the service and deployment (or ds, or whatever) need to be in the same namespace as the operator for these watches to work the way you want. Cross-namespace watches and controller references are not supported in Kubernetes.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,564 @@
...
+			if err = r.client.Update(context.TODO(), instance); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Halt the reconciliation.
+		return reconcile.Result{}, nil
+	}
+
+	// List ConfigMaps with SSH keys
```
lisa, line 181: I'd like to see a comment in this part of the code (dealing with keys) that reminds the reader that changes in the configmaps will be reconciled at least once a minute, so a created or deleted key will have that amount of lag time to be noticed by the deployment. The comment could also remind the reader how that process will work (see my comment about spec equality).

pkg/controller/sshd/sshd_controller.go
lisa: Separately, I'm curious why this approach was taken instead of a single ConfigMap with many keys inside of it used to represent each associate's key, eg:

```yaml
data:
  alice: ssh-rsa ...
  bob: ssh-rsa ...
  barack: ssh-rsa ...
```

It seems like this would be a cleaner effect and the be less work to build up a new list each time during reconciliation and has the added benefit that when mounted inside the container, it would be [kept up to date by Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically). (Though I still like the idea of a separate custom resource for each associate; this single ConfigMap has merit as well.)

What do you think Matt?

go.mod
lisa: (and don't forget `go mod tidy` after)

deploy/80_rh-ssh.Namespace.yaml
mwoodson: I don't think the double curly brace "{{" are needed here.  I think you want single curly brace.

We use double curly brace for the ALLOWED_CIDR_BLOCKS var because this is a list, and the double curly brace expands this into an actual list.  The single curly brace should expand into a string.

deploy/90_rh-ssh.SSHD.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: cloudingress.managed.openshift.io/v1alpha1
+kind: SSHD
+metadata:
+  name: rh-ssh
+  namespace: "${{SSHD_NAMESPACE}}"
```
mwoodson, line 5: same here with the double curly brace

deploy/90_rh-ssh.SSHD.yaml
```diff
@@ -0,0 +1,9 @@
...
+spec:
+  dnsName: rh-ssh
+  allowedCIDRBlocks: "${{ALLOWED_CIDR_BLOCKS}}"
+  image: "${{SSHD_IMAGE}}"
```
mwoodson, line 9: same here with double curly...

pkg/controller/sshd/sshd_controller.go
cblecker: -1 to a CRD for each associate. We don't need to spend a watch every individual associate, we don't need to maintain fields beyond the key, and the only use for this at this point is to drop it into the FS of the container. A config map or secret meets all of these requirements without a new custom data type.

Is there any particular benefit to splitting up into individual files for every associate, as opposed to one authorized_keys file per configmap? The only reason we have to split these into separate configmaps in the first place, is that not all associates get access to all clusters (e.g. CSSRE)

pkg/controller/sshd/sshd_controller.go
cblecker: This is typically done when the operator itself has a configuration change, and needs to update the already installed deployment with a new option.

pkg/controller/sshd/sshd_controller.go
cblecker: +1 to this being a master workload

deploy/80_rh-ssh.Namespace.yaml
mbarnes: We don't _need_ the variable, but the Jenkins jobs @luis-falcon is setting up to generate the SSH key configmaps will need to know what namespace to deposit them.  So my thought was just to keep that name, if not in one place, then at least in one repo (app-interface).

Ack on the single curly brace.  ¯\_(ツ)_/¯

pkg/controller/sshd/sshd_controller.go
mbarnes: The intention between @luis-falcon and I was to generate one configmap and selectorsyncset per team.  Jenkins jobs will generate these objects on Hive.  The configmap will contain all the SSH keys for that team and the selectorsyncset will target clusters appropriate for that team.  (e.g. everything for SREP, but only RHMI clusters for CS-SRE, etc.)

The controller will find whatever teams' configmaps are present on the cluster and prepare a volume/volumemount for each in the pod spec of the deployment.

pkg/controller/sshd/sshd_controller.go
mbarnes: Oh nice!  Didn't know about that package.

pkg/controller/sshd/sshd_controller.go
mbarnes: Right.  I think I've ensured that in the code.

pkg/controller/sshd/sshd_controller.go
mbarnes: Good point.  It's also the case that a _new_ configmap (for a new team) will eventually be detected and an updated deployment created.  I'm not explicitly watching for this event (since there's no ownership linkage to the operator), but just relying on the default reconcile frequency.

deploy/90_rh-ssh.SSHD.yaml
mbarnes: I defer to @cblecker and @mwoodson's expertise on this point.  No strong opinion from me.

mbarnes:
Thanks @lisa, @cblecker, and @mwoodson for the detailed review!  I'll follow up on Monday.

pkg/controller/sshd/sshd_controller.go
lisa: I can see that, @cblecker, but what is expected to be changed at this point, and why?

deploy/80_rh-ssh.Namespace.yaml
mbarnes: By the way, what software is performing the variable substitution here? In other words, where are these curly brace rules coming from? Are they meant to be Jinja templates or something else?

mbarnes:
I've temporarily inserted a couple "**fixup**" commits with changes to address the review comments.
I _think_ I've covered everything except for outstanding questions around how to deploy this.

deploy/90_rh-ssh.SSHD.yaml
mbarnes: I think it makes a lot of sense to at least deploy the namespace through MCC, since it will need a service load balancer quota exemption.  Opened https://github.com/openshift/managed-cluster-config/pull/410 for this.

mbarnes:
Reapplied the **work-in-progress** label so I can handle the fixup commits before this merges.

pkg/controller/sshd/sshd_controller.go
mbarnes: Talked this over with @cblecker and decided we'll just have to do without the watches.
The reconcile loop will still run periodically, and that's good enough for this controller.

mbarnes:
I added a **fixup** commit for the `DeleteARecord` method after noticing the finalizer cycles forever if it can't find the DNS entry it's trying to delete.  The fix makes the method idempotent.

pkg/controller/sshd/sshd_controller.go
cblecker: We don't know yet? We know this this operator owns this object, why wouldn't we enforce the spec of it?

mbarnes:
Made a few deployment changes to hopefully get this in faster:

- Eliminated the `80_rh-ssh.Namespace.yaml` file in favor of moving it to MCC (https://github.com/openshift/managed-cluster-config/pull/410).
- Hard-coded the namespace and image in `90_rh-ssh.SSHD.yaml`, leaving only `ALLOWED_CIDR_BLOCKS` as a variable.  We can revisit this later if necessary.

pkg/awsclient/route53_helper.go
```diff
@@ -1,38 +1,44 @@
...
-			// In order to get the publicHostedZoneID we need to get
-			// the HostedZone.Id object which is in the form of "/hostedzone/Z1P3C0HZA40C0N"
-			// Since we only care about the ID number, we take index of the last "/" char and parse right
-			zoneID := aws.StringValue(zone.Id)
-			slashIndex := strings.LastIndex(zoneID, "/")
-			publicHostedZoneID = zoneID[slashIndex+1:]
+			// The zone ID is the last element of the string
+			// HostedZone.Id, which takes the form of a path:
+			// "/hostedzone/<ZONEID>"
+			publicHostedZoneID = path.Base(aws.StringValue(zone.Id))
```
lisa, line 27: I'm not sure about this change with using path package in this way. Path seems to want to be used with files and URLs. Is it coincidence that these hosted zones take that form? There's no doubt that it works but it's a counterintuitive use, eh?

pkg/controller/sshd/sshd_controller.go
lisa: I want to revise this:

It seems like it should work as is so long as the CR is in the same namespace as the objects being watched. In this case, the CR will be in `openshift-sre-sshd` and the child resources will also be in the same namespace. If the CR was in the same namespace where the operator runs (`openshift-cloud-ingress-operator`), I think you'd hit these problems.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,580 @@
...
+
+	err = r.ensureDNSRecords()
+	if err != nil {
+		r.SetSSHDStatusError(instance, "Failed to ensure the DNS record", err)
+		return reconcile.Result{}, err
+	}
+
+	r.SetSSHDStatus(instance, "SSHD is ready", cloudingressv1alpha1.SSHDStateReady)
+
+	return reconcile.Result{RequeueAfter: 60 * time.Second}, nil
```
lisa, line 344: I'm not sure you need to reconcile every minute like the apischeme controller because your dependent watches should cause reconciliation upon their change.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,580 @@
...
+
+func getMatchLabels(cr *cloudingressv1alpha1.SSHD) map[string]string {
+	return map[string]string{"deployment": cr.Name}
+}
+
+func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.ConfigMapList) *appsv1.Deployment {
+	// Prefer nil over empty slices to satisfy reflect.DeepEqual.
+	var volumes []corev1.Volume
+	var volumeMounts []corev1.VolumeMount
+	volumeDefaultMode := int32(0600)
```
lisa, line 355: Another use case for `pointer` utils.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -0,0 +1,580 @@
...
+				Name:          "ssh",
+				ContainerPort: int32(2222),
+				Protocol:      corev1.ProtocolTCP,
+			},
+		},
+		VolumeMounts:             volumeMounts,
+		TerminationMessagePath:   "/dev/termination-log",
+		TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+		ImagePullPolicy:          corev1.PullAlways,
+	}
```
lisa, line 391: A few questions here:

1. Why separate this out in its own variable and not define it in-line?
2. Where is the image pipeline?
3. In that image what is `/opt/start-sshd.sh`? And will it be clear to folks who wish to use an "alternate" image that they must implement that?

pkg/awsclient/route53_helper.go
mbarnes: As I understand it, [path.filepath](https://golang.org/pkg/path/filepath/) is for files but [path](https://golang.org/pkg/path/filepath/) is for any kind of slash-separated identifier.

Hosted zone IDs _happen_ to be path-like, and as long as we're using that knowledge to pick it apart, [path](https://golang.org/pkg/path/filepath/) seems to me like an appropriate use.

pkg/controller/sshd/sshd_controller.go
mbarnes: :+1: Oversight on my part.

pkg/controller/sshd/sshd_controller.go
mbarnes: 1. Just for easier readability (imo).  Large sprawling struct literals are easy to get lost in.
2. Pipeline is on [ci.ext](https://ci.ext.devshift.net/view/sre-ssh-proxy/) and the source repo is [sre-ssh-proxy-container](https://github.com/openshift/sre-ssh-proxy-container/).
3. Startup script is [here](https://github.com/openshift/sre-ssh-proxy-container/blob/master/container/start-sshd.sh), which involves some gymnastics to allow it to run as a non-privileged pod.  @mwoodson [documented](https://docs.google.com/document/d/1guvdWeFw-KjZQSc3dv3udTQdP7nfARQuTy1Ya7kOH3A/edit#) the complexities in his proof-of-concept.  The only "alternate" images I anticipate would be test images built from a fork of the source repo.

pkg/controller/sshd/sshd_controller.go
mbarnes: Yes, that's the plan.  The CR and its child resources will be in `openshift-sre-sshd`.

pkg/controller/sshd/sshd_controller.go
mbarnes: As I've come to understand it, this idiom both applies _intentional_ changes to the spec but also reverts _unintentional_ changes from, for example, attempted manual edits to the resource.

pkg/awsclient/route53_helper.go
lisa: The note from the `path` [package docs](https://pkg.go.dev/path?tab=doc):

> The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the [path/filepath](https://pkg.go.dev/path/filepath@go1.14.6?tab=doc) package.

I don't think this is a particularly important hill for me to die on, but I want to be kind to our future selves with idiomatic use of language packages. Maybe someone with more experienced Go chops than me could weigh in? @cblecker maybe?

pkg/controller/sshd/sshd_controller.go
lisa: I'd add a note here that it's strange for an operator to create objects in distant namespaces, but that's literally what this entire operator is doing before this PR even came to be. ;-) But hey, it's cluster-scoped, so let's roll with it.

pkg/controller/sshd/sshd_controller.go
lisa: Thanks! I minorly disagree with the first point, but 👍 

Are there thoughts to the shifting ssh host key being stored so that for each cluster it's more or less static? That topic is kind of out of scope for this PR but I want to make sure that engineers who are connecting to clusters don't have to worry that they're connecting to a potentially different system/host key.

pkg/controller/sshd/sshd_controller.go
lisa: @mbarnes Not seeing controls here to direct this workload on the master nodes? Missing the `nodeAffinity` portion of the spec?

pkg/controller/sshd/sshd_controller.go
mbarnes: cc @mwoodson
That's a valid concern and something I encountered while testing the container.  I kept having to delete the previous container's host key from my `known_hosts` file to keep `ssh` quiet.  Possibly a set of static per-cluster host keys could be mounted in from a ConfigMap, but I'm unsure about the Hive logistics.

lisa: > I kept having to delete the previous container's host key from my `known_hosts` file to keep `ssh` quiet.

This is fine during testing, but in production can lead to dangerous habits. :( I'd love to see a "save the key" solution. Curious if @jaybeeunix has an opinion?

jaybeeunix: I agree with @lisa's want for static keys for the same reason. If we can't do that, what ways can we reduce the bad behavior of ignoring known_hosts issues? (maybe something educational and/or procedural?)

Some future feature to this might be to use PKI, which would also eliminate the bad habit of saying "yes" on the first connection.

mwoodson: I agree with the concern as well.

Do we feel strongly this should block the initial deployment, or can this be an enhancement that we work with next?  I would vote for making this an enhancement.

jaybeeunix: I'm fine with treating it as an enhancement, if we have something educational/procedural to reduce the bad habits.

If we're looking at it in the future, then I'd even skip the static keys and explore PKI fully (as static keys don't address the other bad habit of saying yes on the initial connection without verifying the fingerprint).

pkg/controller/sshd/sshd_controller.go
mbarnes: I used a `nodeSelector`.  Would `nodeAffinity` be preferred?

pkg/awsclient/route53_helper.go
cblecker: Amazon's Route 53 API for making changes to hosted zones is designed as a restful API. The ID [returned here](https://docs.aws.amazon.com/Route53/latest/APIReference/API_HostedZone.html) is a URI path, and IMO is an appropriate use of the path package based on:
> The path package should only be used for paths separated by forward slashes, such as the paths in URLs.

lisa: Works for me 👍

sam-nguyen7: +1 this looks a lot cleaner than before. Since we are using it in the other controllers though it would be good to test and make sure everything works as expected

pkg/controller/sshd/sshd_controller.go
mbarnes: Added a commit to use `nodeAffinity`.

pkg/controller/sshd/sshd_controller.go
mbarnes: Note added.

pkg/awsclient/route53_helper.go
mbarnes: I've tested it in the context of the SSHD controller and it works fine afaict.

pkg/controller/sshd/sshd_controller.go
mbarnes: For the record, @mwoodson opened [OSD-4672](https://issues.redhat.com/browse/OSD-4672) as a follow-up spike.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -415,9 +415,24 @@ func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.Conf
 					RestartPolicy:                 corev1.RestartPolicyAlways,
 					TerminationGracePeriodSeconds: pointer.Int64Ptr(30),
 					DNSPolicy:                     corev1.DNSClusterFirst,
-					NodeSelector:                  map[string]string{nodeMasterLabel: ""},
 					SecurityContext:               &corev1.PodSecurityContext{},
 					SchedulerName:                 "default-scheduler",
+					Affinity: &corev1.Affinity{
+						NodeAffinity: &corev1.NodeAffinity{
+							RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
```
cblecker, line 429:
```suggestion
							PreferredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
```

deploy/90_rh-ssh.SSHD.yaml
csheremeta: This will work to pull in the IPs of the bastion nodes and the hive v4 shard NAT(s). If other IPs need to be included (i.e. the v3 hive node IPs, we'd need to pull those from the APIScheme object

lisa:
@mbarnes Looks good, only lacking unit tests now. 👍

deploy/90_rh-ssh.SSHD.yaml
mbarnes: > Great! I think this should also be single curlies, @mbarnes

Ack.  So _all_ variables should be single curlies?

mbarnes:
Test file added.  Just simple unit tests on the child resources for now.  Didn't have time for any interface mocking.

I can revisit this to add more extensive testing at a later date.

sam-nguyen7:
/lgtm 

this is good for first iteration of unit tests

csheremeta:
Matthew gave us a demo of this this morning. I had an issue with my SSH keys, so I couldn't get into the ssh pod. What he showed us lgtm, though
/hold cancel

PR #77, opened 2020-07-15
Add Router Service Controller And Watcher For Annotations

dustman9000:
These are updates to support setting the `service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout=600` annotation on the LoadBalancer services in the openshift-ingress namespace to provide a short-term solution for [OSD-3944](https://issues.redhat.com/browse/OSD-3944).

### DIFF
--- a/pkg/controller/add_router_service.go
+++ b/pkg/controller/add_router_service.go
@@ -1,0 +1,10 @@
+package controller
+
+import (
+	"github.com/openshift/cloud-ingress-operator/pkg/controller/routerservice"
+)
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	AddToManagerFuncs = append(AddToManagerFuncs, routerservice.Add)
+}
--- a/pkg/controller/routerservice/router_service_controller.go
+++ b/pkg/controller/routerservice/router_service_controller.go
@@ -1,0 +1,117 @@
+package routerservice
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_router_service")
+
+const (
+	RouterServiceNamespace = "openshift-ingress"
+	ELBAnnotationKey       = "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout"
+	ELBAnnotationValue     = "600"
+)
+
+// Add creates a new Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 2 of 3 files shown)


pkg/controller/service/service_controller.go
```diff
@@ -0,0 +1,125 @@
+package service
```
mwoodson, line 1: A more specific name might be better?  router_service_controller.go?

pkg/controller/service/service_controller.go
```diff
@@ -0,0 +1,125 @@
...
+const (
+	SERVICE_NAMESPACE    = "openshift-ingress"
+	ELB_ANNOTATION_KEY   = "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout"
+	ELB_ANNOTATION_VALUE = "600"
+)
+
+/**
+* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
+* business logic.  Delete these comments after modifying this file.*
+ */
```
mwoodson, line 32: delete these comments.

pkg/controller/service/service_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_service")
+
+const (
+	SERVICE_NAMESPACE    = "openshift-ingress"
```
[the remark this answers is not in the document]
lisa: +1 `ServiceNamespace` would be in line with Go conventions.

pkg/controller/service/service_controller.go
dustman9000: Done. I renamed to package, file names, and reconcile function.

pkg/controller/service/service_controller.go
dustman9000: done.

pkg/controller/service/service_controller.go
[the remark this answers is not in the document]
dustman9000: done.

pkg/controller/service/service_controller.go
```diff
@@ -0,0 +1,125 @@
...
+			err = r.client.Update(context.TODO(), svc)
+			if err != nil {
+				reqLogger.Error(err, "Error updating service annotation")
+				return reconcile.Result{}, err
+			}
+		} else {
+			reqLogger.Info("skipping service " + svc.Name + " w/ proper annotations")
+		}
+	} else {
+		reqLogger.Info("ignoring service: " + svc.Name)
```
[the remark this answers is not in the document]
dustman9000: Removed excess logging.

dustman9000:
> @dustman9000 @mwoodson do we want to set this for every ELB or should it be configurable? I don't think we should just blanket set all ELBs to 600 seconds as the change could be unexpected by some customers. Thoughts?

We are only doing this for the ELBs (LoadBalancer types) managed by ingresscontroller. These are only apps and apps2. The API and RH-API ELBs are left untouched. The collective decision on OSD-3944 was to make this fleet-wide of 600s until the RFE is implemented.

mwoodson:
I think making it configurable is more work than we want to do at this point.  I'm -1 to making it config, and setting it at 10 minutes across the board.

In talking to Ben Bennet, he believes the HA Proxy timeout setting is set at 5 minutes for http and https calls, websockets are at 1 hour.  So even if we have 10 minutes (which i think we should keep), ha proxy will kill terminate connections at 5 minutes.

blanket to 10 minutes is what we did in v3.  We didn't have any complaints that I remember.

lisa:
Is the plan for this to be removed once the RFE is adopted?

Do we need a plan to backfill?

lisa:
/hold 

for review plz

jharrington22:
If v3 had 10 minutes across the board without issue and HAProxy is < 10 minutes then I'm +1 to this across the board.

A configurable value in a ConfigMap would be ideal but out of scope for this PR current value is 60 seconds across the board so I say we card up the work for the ConfigMap and only complete it if we have a requirement from the customer or the RFE doesn't get accepted.

PR #78, opened 2020-07-23
Add APIScheme object back to CIO for v4 hive sharding purposes

csheremeta:


deploy/90_rh-api.APIScheme.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: cloudingress.managed.openshift.io/v1alpha1
+kind: APIScheme
+metadata:
+  name: rh-api
+  namespace: openshift-cloud-ingress-operator
+  annotations: {
```
jharrington22, line 6: no need for the `{` its yaml :)

PR #81, opened 2020-07-27
Add debug logs

sam-nguyen7:
I'm primarily concerned with 
1. What the initial list of ingresscontroller looks like on cluster
2. What we have in the publshingstrategy
3. The list of the difference. aka what we have in publishingstrategy that is NOT on cluster. - these are the ones that the reconcile should action on.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -139,6 +139,12 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		}
 	}
 
+	// ----------------------------TODO: remove these debug logs afterwards --------------------------------------
+	log.Info(fmt.Sprintf("initial ingresscontroller list: %v", ingressControllerList.Items))
+	log.Info(fmt.Sprintf("appingress on publishingstrategy CR: %v", instance.Spec.ApplicationIngress))
+	log.Info(fmt.Sprintf("ingress to reconcile: %v", ingressNotOnCluster))
+	// -----------------------------------------------------------------------------------------------------------
+
 	for _, appingress := range ingressNotOnCluster {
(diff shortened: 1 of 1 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -139,6 +139,12 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		}
 	}
 
+	// ----------------------------TODO: remove these debug logs afterwards --------------------------------------
+	log.Info(fmt.Sprintf("initial ingresscontroller list: %v", ingressControllerList.Items))
+	log.Info(fmt.Sprintf("appingress on publishingstrategy CR: %v", instance.Spec.ApplicationIngress))
+	log.Info(fmt.Sprintf("ingress to reconcile: %v", ingressNotOnCluster))
```
lisa, line 145:
```suggestion
	log.Info(fmt.Sprintf("initial ingresscontroller list: %+v", ingressControllerList.Items))
	log.Info(fmt.Sprintf("appingress on publishingstrategy CR: %+v", instance.Spec.ApplicationIngress))
	log.Info(fmt.Sprintf("ingress to reconcile: %+v", ingressNotOnCluster))
```

`%+v` will probably be a bit better to print field names.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -139,6 +139,12 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		}
 	}
 
+	// ----------------------------TODO: remove these debug logs afterwards --------------------------------------
```
2uasimojo, line 142: If you wanted to keep these messages long-term, you could use log.V(some_high-ish_number).Info(...). That way you could switch them on in a given cluster without a code change in the future.

PR #84, opened 2020-08-18
wrap ALLOWED_CIDR_BLOCKS in double braces for array expansion

gregsheremeta:
/cc @lisa 
how on earth do you have that nick
/cc @sam-nguyen7 
/cc @csheremeta


lisa:
/unhold
/lgtm

:shipit:

PR #86, opened 2020-08-18
Use double braces for ENABLE_MANAGEMENT_API_SERVER_INGRESS

2uasimojo:
...which was showing up as a stringified boolean rather than a boolean.


lisa:
@2uasimojo This PR doesn't seem to interact with `ENABLE_MANAGEMENT_API_SERVER_INGRESS`, but rather `allowedCIDRBlocks`? The subject of this PR would imply something different.

PR #88, opened 2020-08-20
Ensure ingresscontrollers matches publishingstrategy

sam-nguyen7:
Add check to ensure that expected ingresscontrollers are present on cluster. If we fail to do that, requeue until successful

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -160,6 +160,23 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		}
 	}
 
+	ingressList := &operatorv1.IngressControllerList{}
+	listOptions = []client.ListOption{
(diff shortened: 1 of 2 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -352,6 +359,30 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+				break
+			}
+			listening := string(app.Listening)
+			capListening := strings.Title(strings.ToLower(listening))
+			if ingress.Spec.Domain == app.DNSName && capListening == string(ingress.Status.EndpointPublishingStrategy.LoadBalancer.Scope) {
+				exists = true
+			}
+		}
+		if !exists {
+			isContained = false
```
mwoodson, line 380: Should there be a break here?

I would think that if it's false 1x, we want to return false, not keep going.

sam-nguyen7: +1 added break statement

PR #91, opened 2020-09-01
sshd: Generate host keys and bundle into a secret 

mbarnes:
Generating only a 4096-bit RSA host key for now.  Not sure if we want other key types as well like ED25519.

Also added environment variables to the pod for various mount points, hoping I can teach the SSHD image to generate an `sshd_config` from a template at startup.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -306,7 +306,7 @@ objects:
           nodeAffinity:
             preferredDuringSchedulingIgnoredDuringExecution:
             - preference:
-              matchExpressions:
+                matchExpressions:
                 - key: node-role.kubernetes.io/infra
                   operator: Exists
               weight: 1
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -359,7 +359,7 @@ objects:
       spec:
         dnsName: rh-ssh
         allowedCIDRBlocks: ${{ALLOWED_CIDR_BLOCKS}}
-        image: quay.io/app-sre/sre-ssh-proxy:v44-262ce16
+        image: quay.io/app-sre/sre-ssh-proxy:v49-1fde8e5
 - apiVersion: hive.openshift.io/v1
   kind: SelectorSyncSet
   metadata:
--- a/pkg/controller/sshd/sshd_controller.go
+++ b/pkg/controller/sshd/sshd_controller.go
@@ -9,6 +9,12 @@ import (
 	"strings"
 	"time"
 
+	// For host key generation
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+
 	cloudingressv1alpha1 "github.com/openshift/cloud-ingress-operator/pkg/apis/cloudingress/v1alpha1"
 	"github.com/openshift/cloud-ingress-operator/pkg/awsclient"
 	"github.com/openshift/cloud-ingress-operator/pkg/config"
@@ -21,6 +27,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -112,6 +119,7 @@ type ReconcileSSHD struct {
 
 const (
 	authorizedKeysMountPath   = "/var/run/authorized_keys.d"
+	hostKeysMountPath         = "/var/run/ssh"
 	nodeMasterLabel           = "node-role.kubernetes.io/master"
 	reconcileSSHDFinalizerDNS = "dns.cloudingress.managed.openshift.io"
 )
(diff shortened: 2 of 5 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -306,7 +306,7 @@ objects:
           nodeAffinity:
             preferredDuringSchedulingIgnoredDuringExecution:
             - preference:
-              matchExpressions:
+                matchExpressions:
```
lisa, line 309: What happened here for this change?

pkg/controller/sshd/sshd_controller.go
```diff
@@ -355,7 +395,41 @@ func getMatchLabels(cr *cloudingressv1alpha1.SSHD) map[string]string {
...
+func newSSHDSecret(namespace, name string) (*corev1.Secret, error) {
+	// Generate 2048-bit RSA key
+	rsaKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, err
+	}
+	ssh_host_rsa_key := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(rsaKey),
+	})
```
lisa, line 407: @jaybeeunix Security 👀 ?

pkg/controller/sshd/sshd_controller.go
```diff
@@ -378,6 +452,45 @@ func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.Conf
...
+		return volumeMounts[i].Name < volumeMounts[j].Name
+	})
+
+	volumeName := hostKeysSecret.ObjectMeta.Name
+	volumes = append(volumes, corev1.Volume{
+		Name: volumeName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName:  hostKeysSecret.ObjectMeta.Name,
+				DefaultMode: pointer.Int32Ptr(0600),
```
lisa, line 469:
```suggestion
				DefaultMode: pointer.Int32Ptr(0400),
```
I think you want `0400` to prevent writes inside the pod.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -378,6 +452,45 @@ func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.Conf
...
+	env = append(env, corev1.EnvVar{
+		Name:  "AUTHORIZED_KEYS_DIR",
+		Value: authorizedKeysMountPath,
+	})
+	for key := range hostKeysSecret.Data {
+		env = append(env, corev1.EnvVar{
+			Name:  strings.ToUpper(key),
+			Value: filepath.Join(hostKeysMountPath, key),
+		})
+	}
```
lisa, line 490: What's happening here? There's only one element in the secret's data (cf lines 424-428), so why iterate over it like this instead of a direct reference to that single element?

hack/olm-registry/olm-artifacts-template.yaml
mbarnes: I'm not sure?  I ran `make` and found that.

pkg/controller/sshd/sshd_controller.go
mbarnes: Note, 2048 bits matches [ssh-keygen](https://linux.die.net/man/1/ssh-keygen)'s default for RSA.
> -b bits
Specifies the number of bits in the key to create. For RSA keys, the minimum size is 768 bits and the default is 2048 bits. Generally, 2048 bits is considered sufficient. DSA keys must be exactly 1024 bits as specified by FIPS 186-2.

mbarnes: Note, 2048 bits matches [ssh-keygen](https://linux.die.net/man/1/ssh-keygen)'s default for RSA.
> -b bits
Specifies the number of bits in the key to create. For RSA keys, the minimum size is 768 bits and the default is 2048 bits. Generally, 2048 bits is considered sufficient. DSA keys must be exactly 1024 bits as specified by FIPS 186-2.

hack/olm-registry/olm-artifacts-template.yaml
mbarnes: I'm not sure?  I ran `make` and found this.

pkg/controller/sshd/sshd_controller.go
mbarnes: On the possibility that we might [support other key types](https://github.com/openshift/cloud-ingress-operator/pull/91/files#diff-e5ea7d858af86730d41fd21b37ca353aR409-R413).  Just an attempt at future proofing.

```
// XXX Generate other key types?
//     ECDSA:   Easy, fully supported in standard library.
//     ED25519: Standard library can generate a private key, but requires an
//              external module to create an "OPENSSH PRIVATE KEY" PEM block:
//              https://github.com/mikesmitty/edkey
```

lisa: Ah, I see. Be a little cautious of premature optimizing. :)

hack/olm-registry/olm-artifacts-template.yaml
lisa: Spooky.

pkg/controller/sshd/sshd_controller.go
jaybeeunix: for ref, nist tends to recommend >= 3072 rsa for security needs through 2030, and my openssh on f31 (openssh-8.1p1-1.fc31.x86_64) defaults to 3072 (from the same manpage you mentioned).

it looks like sre-p have said we want 4096 rsa keys: https://github.com/openshift/ops-sop/blob/master/v3/operationsorientation.asciidoc#62-ssh-keys , so let's go with 4096.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -378,6 +452,45 @@ func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.Conf
...
+		return volumeMounts[i].Name < volumeMounts[j].Name
+	})
+
+	volumeName := hostKeysSecret.ObjectMeta.Name
+	volumes = append(volumes, corev1.Volume{
+		Name: volumeName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName:  hostKeysSecret.ObjectMeta.Name,
+				DefaultMode: pointer.Int32Ptr(0400),
```
cblecker, line 469: It's worthwhile to put a comment on why you're setting this specific mode

cblecker (commented):
A couple non-blocking comments, otherwise lgtm

pkg/controller/sshd/sshd_controller.go
mbarnes: Done, amended the original commit.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -355,7 +395,41 @@ func getMatchLabels(cr *cloudingressv1alpha1.SSHD) map[string]string {
 	return map[string]string{"deployment": cr.Name}
 }
 
-func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.ConfigMapList) *appsv1.Deployment {
+func newSSHDSecret(namespace, name string) (*corev1.Secret, error) {
```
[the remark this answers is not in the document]
mbarnes: Added unit test in a new commit.

cblecker:
/lgtm

/assign @lisa 
Assigning Lisa for approval

lisa (approved):
/lgtm

PR #93, opened 2020-09-08
sshd: Make sure service annotation is applied

mbarnes:
This will fix SSHD-controlled load balancer services that are missing the LB connection idle timeout annotation.

(Noticed on some production clusters, since the annotation was added shortly after the initial feature promotion.)

### DIFF
--- a/pkg/controller/sshd/sshd_controller.go
+++ b/pkg/controller/sshd/sshd_controller.go
@@ -122,6 +122,8 @@ const (
 	hostKeysMountPath         = "/var/run/ssh"
 	nodeMasterLabel           = "node-role.kubernetes.io/master"
 	reconcileSSHDFinalizerDNS = "dns.cloudingress.managed.openshift.io"
+	ELBAnnotationKey          = "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout"
+	ELBAnnotationValue        = "600"
 )
 
 // Reconcile reads that state of the cluster for a SSHD object and makes changes based on the state read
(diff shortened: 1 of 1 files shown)


pkg/controller/sshd/sshd_controller.go
```diff
@@ -324,15 +326,28 @@ func (r *ReconcileSSHD) Reconcile(request reconcile.Request) (reconcile.Result,
...
 		service.Spec.HealthCheckNodePort = foundService.Spec.HealthCheckNodePort
 		if !reflect.DeepEqual(foundService.Spec, service.Spec) {
-			// Specs aren't equal, update and fix.
 			r.SetSSHDStatusPending(instance, "Updating service", "from", foundService.Spec, "to", service.Spec)
 			foundService.Spec = *service.Spec.DeepCopy()
+			serviceNeedsUpdate = true
+		}
+
+		if serviceNeedsUpdate {
 			if err = r.client.Update(context.TODO(), foundService); err != nil {
```
lisa, line 351: Is this needed? If the `reflect.DeepEqual` is false, then doesn't it by definition need an update?

mbarnes: It does, but a missing annotation (just above) also sets `serviceNeedsUpdate = true`.

lisa: I see what's happening now. Note to past me: Don't review code when tired!

lisa (commented):
/lgtm

PR #96, opened 2020-09-22
sshd: Explicitly make host keys group readable

mbarnes:
The sshd process runs as a non-root user within the pod, whereas the mounted host keys are owned by root.  We therefore depend on the key files being group-readable.  Set this read bit explicitly instead of depending on the security context to set it.

Normally group-readable private keys are forbidden by OpenSSH, but it turns out the server [does NOT apply its strict permission checks when a private key is owned by a different user than itself](https://github.com/openssh/openssh-portable/blob/9bb8a303ce05ff13fb421de991b495930be103c3/authfile.c#L96-L104).

### DIFF
--- a/pkg/controller/sshd/sshd_controller.go
+++ b/pkg/controller/sshd/sshd_controller.go
@@ -475,20 +475,24 @@ func newSSHDDeployment(cr *cloudingressv1alpha1.SSHD, configMapList *corev1.Conf
 		return volumeMounts[i].Name < volumeMounts[j].Name
 	})
 
+	// Because the OpenSSH server runs as an arbitrary user instead of root,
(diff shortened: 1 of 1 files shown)


lisa:
@mbarnes What are the owner:group owner of the file inside the container expected to be?

mwoodson (approved):
/lgtm

georgettica (commented):
If that was tested I can give a thumbs up
/approve

mbarnes:
> If that was tested I can give a thumbs up

It was tested.  Used [BZ 1868976#c25](https://bugzilla.redhat.com/show_bug.cgi?id=1868976#c25) as a reproducer.

PR #97, opened 2020-09-24
[WIP] apischeme: Adapt reconcile to use cloudclient package

mbarnes:
**Note:** This PR is against the **cloudclient** branch, not **master**.

The **cloudclient** branch at this point just adds the `cloudclient` package.  It was split off so the team could work on other controllers in parallel while at the same time ironing any kinks out of the interface and AWS provider.

Once I can demonstrate the interface working by way of the APIScheme controller, I'll be comfortable merging the `cloudclient` package to the **master** branch.

* 1st commit (`"apischeme: Adapt reconcile to use cloudclient package"`) is taken verbatim from the [OSD-3566](https://github.com/lisa/cloud-ingress-operator/commit/8c65bebd70c3d392f6306f11400e0f8942c6a2a3#diff-790e8873bdd0a7774c76cb2579f89c86) branch.
* 2nd commit (`"apischeme: Implement a finalizer for the DNS entries"`) are my own additions.


lisa:
@mbarnes Is this still a work in progress? I see the associated Jira card is in "code review" phase.

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -118,9 +123,73 @@ func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Res
...
+
+	// Check for a deletion timestamp.
+	if instance.DeletionTimestamp.IsZero() {
+		// Request object is alive, so ensure it has the DNS finalizer.
+		if !controllerutil.ContainsFinalizer(instance, reconcileFinalizerDNS) {
+			controllerutil.AddFinalizer(instance, reconcileFinalizerDNS)
+			if err = r.client.Update(context.TODO(), instance); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
```
lisa, line 149: Upon success, should this cause the object to requeue the Reconcile again, since it has been updated? If so, be sure to adjust the comment, and requeue.

lisa (changes requested):
Assuming this isn't still a work in progress, here is a review in part.

pkg/controller/apischeme/apischeme_controller.go
mbarnes: Is it necessary to requeue immediately after an update?  (I honestly don't know.)

lisa: @mbarnes Yes. The reason is: When you write the object back to the API server, Kubernetes updates the version in the metadata and further writes to the original object will fail because the "source" object is older than what the API server knows about. This is to protect against overwriting, or time travelling to before previous saves.

General best practice is to do the write and return out of the Reconcile method to requeue again.

mbarnes: That makes a lot of sense.  Thanks for the detailed explanation!

mbarnes: On re-reading the logic, it returns anyway right after the update, so there's no subsequent updates.  The next reconciliation pass _should_ be a no-op, but I requeued anyway.  I guess it can't hurt.

mbarnes:
Merging this into `cloudclient` with @lisa's suggestion implemented.
Bit more testing and I'll submit a PR to merge this much into `master`.

PR #98, opened 2020-09-29
OSD-5267 cloudclient sshd controller refactor

yithian:


lisa:
How does this line up with #97? Are the cloudclient code the same?

yithian:
yeah, the cloudclient code will be the same, but i realize now that i rebased this to the master branch instead of the cloudclient branch...

/hold

PR #99, opened 2020-09-29
OSD-5267 cloudclient sshd controller refactor

yithian:


pkg/controller/sshd/sshd_controller.go
```diff
@@ -303,44 +295,7 @@ func (r *ReconcileSSHD) Reconcile(request reconcile.Request) (reconcile.Result,
...
-	exists, elb, err := r.awsClient.DoesELBExist(elbName)
-	if err != nil {
-		r.SetSSHDStatusError(instance, "Failed to get load balancer status", err)
-		return reconcile.Result{}, err
-	}
-	if !exists {
-		// It isn't bad that it doesn't exist if there's no error, so re-queue.
-		r.SetSSHDStatusPending(instance, "Waiting on service load balancers")
-		return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
-	}
```
yithian, line 331: I didn't see any functionality to check if an ELB exists in cloudclient. Is this a gap in the reconcile loop, now?

lisa: @yithian Refer to https://github.com/openshift/cloud-ingress-operator/blob/3a741b31f9c91bada491f81d512b407d8e148cab/pkg/cloudclient/aws/private.go#L180-L200 in the aws package.

The controller shouldn't need to know anything about the implementation because all it wants is for the DNS entry to be present (or absent) for the associated `Service`. It will call the appropriate interface method to set/delete that record.

However, because it's a common thing that creating the backing cloud load balancer takes time, there is a [bespoke error](https://github.com/openshift/cloud-ingress-operator/blob/3a741b31f9c91bada491f81d512b407d8e148cab/pkg/errors/errors.go#L7-L9) which can be returned by the cloud provider packages to indicate that the ELB (in AWS) isn't created yet. [The controller can make use of that returned error](https://github.com/openshift/cloud-ingress-operator/blob/3a741b31f9c91bada491f81d512b407d8e148cab/pkg/controller/apischeme/apischeme_controller.go#L225-L246) to take some relevant action, such as requeueing in ten seconds.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -630,41 +585,21 @@ func newSSHDService(cr *cloudingressv1alpha1.SSHD) *corev1.Service {
...
-			r.route53.resourceRecordSetName,
-			"RH SSH Endpoint", false)
+		err := r.cloudClient.EnsureSSHDNS(context.TODO(), r.client, cr, svc)
 		if err != nil {
-			log.Info("Couldn't upsert a DNS record for public zone: " + err.Error())
+			log.Info("Couldn't ensure DNS record for zone: " + err.Error())
 			if i == config.MaxAPIRetries {
-				log.Info("Out of retries for public zone")
+				log.Info("Out of retries for private zone")
 				return err
```
mbarnes, line 603: I had assumed the retry loop for "ensure" and "delete" DNS methods had been moved into `CloudClient` but it appears not.  @lisa: Should they be?  Strikes me as implementation detail controllers shouldn't have to deal with.

mbarnes, line 603: The retry loop for "ensure" and "delete" DNS records has been moved into `pkg/cloudclient/aws/private.go`, so this function can be simplified and maybe eliminated entirely.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -677,41 +612,21 @@ func (r *ReconcileSSHD) ensureDNSRecords() error {
...
-			r.route53.loadBalancerHostedZoneId,
-			r.route53.resourceRecordSetName,
-			false)
+		err := r.cloudClient.DeleteSSHDNS(context.TODO(), r.client, cr, svc)
 		if err != nil {
-			log.Info("Couldn't delete a DNS record for public zone: " + err.Error())
+			log.Info("Couldn't delete a DNS record: " + err.Error())
 			if i == config.MaxAPIRetries {
-				log.Info("Out of retries for public zone")
+				log.Info("Out of retries for zone")
```
mbarnes, line 629: The retry loop for "ensure" and "delete" DNS records has been moved into `pkg/cloudclient/aws/private.go`, so this function can be simplified and maybe eliminated entirely.

mbarnes (changes requested):

pkg/controller/sshd/sshd_controller.go
yithian: This was super helpful - thanks! I've pushed a commit that should catch that error and requeue.

pkg/controller/sshd/sshd_controller.go
yithian: I've removed this function since, as you said, it's redundant.

pkg/controller/sshd/sshd_controller.go
yithian: I've removed this function since, as you said, it's redundant.

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -107,7 +107,7 @@ func (r *ReconcileAPIScheme) Reconcile(request reconcile.Request) (reconcile.Res
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			reqLogger.Error(err, "Couldn't find the APIScheme object")
+			reqLogger.Info("Couldn't find the APIScheme object")
```
lisa, line 110: Is this change meant to be in this PR?

yithian: ugh, no. it got swept in by rebase conflicts, i think. fixing...

mbarnes (commented):
Looks ok to me, but I'll defer to @lisa as there's some stylistic things I've not seen done before like switching on possibly **nil** error variables.  (I've always seen `if err != nil { do error handling }`.)

Appreciate the additional unit tests.  I'll see if I can steal some of that for APIScheme.

lisa:
> Looks ok to me, but I'll defer to @lisa as there's some stylistic things I've not seen done before like switching on possibly **nil** error variables. (I've always seen `if err != nil { do error handling }`.)
> 

I wanted to be clear: The pattern is to be tested. I'm assuming it works because of the relationship between if statements and switch statements. I think it'll work, but I'm not sure. It compiled anyways. ;-)

yithian:
go's coverage tool shows that the unit test proceeds past the nil option in the switch statement, for what it's worth

pkg/controller/sshd/sshd_controller.go
```diff
@@ -178,9 +169,27 @@ func (r *ReconcileSSHD) Reconcile(request reconcile.Request) (reconcile.Result,
...
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			err = r.cloudClient.DeleteSSHDNS(context.TODO(), r.client, instance, svc)
+			switch err {
+			case nil:
+				// all good
+			case err.(*cioerrors.LoadBalancerNotFoundError):
+				// couldn't find the new ELB yet
```
lisa, line 187: Can you expand this comment with what it means that the ELB isn't present? I would think that during a delete it means that the ELB hasn't been created yet (and there is such a creation wanting to happen).

yithian: That's my understanding too. Maybe it was created and then immediately removed, and might have to wait for the ELB to be created before it can be deleted?

I'm fine removing the error check for `LoadBalancerNotFoundError` on deletion if we don't think it can get into that situation, though.

pkg/controller/sshd/sshd_controller.go
```diff
@@ -358,45 +367,16 @@ func (r *ReconcileSSHD) Reconcile(request reconcile.Request) (reconcile.Result,
...
-	if !exists {
-		// It isn't bad that it doesn't exist if there's no error, so re-queue.
-		r.SetSSHDStatusPending(instance, "Waiting on service load balancers")
+	err = r.cloudClient.EnsureSSHDNS(context.TODO(), r.client, instance, service)
+	switch err {
+	case nil:
+		// all good
+	case err.(*cioerrors.LoadBalancerNotFoundError):
+		// couldn't find the new ELB yet
+		r.SetSSHDStatus(instance, "Couldn't reconcile", "AWS ELB isn't ready yet.")
```
mbarnes, line 376: nit: This should be worded less cloud-specifically.  Maybe: `"Load balancer isn't ready yet"`

pkg/controller/sshd/sshd_controller.go
```diff
@@ -178,9 +169,27 @@ func (r *ReconcileSSHD) Reconcile(request reconcile.Request) (reconcile.Result,
...
+				return reconcile.Result{}, err
+			}
+
+			err = r.cloudClient.DeleteSSHDNS(context.TODO(), r.client, instance, svc)
+			switch err {
+			case nil:
+				// all good
+			case err.(*cioerrors.LoadBalancerNotFoundError):
+				// couldn't find the ELB - it's likely still queued for creation
+				r.SetSSHDStatus(instance, "Couldn't reconcile", "AWS ELB isn't ready.")
```
mbarnes, line 188: nit: This should be worded less cloud-specifically. Maybe: "Load balancer isn't ready yet"

mbarnes (changes requested):
One minor nit in two different places.  Other than that it all looks good to me.  :+1:

mbarnes (approved):

PR #104, opened 2020-10-13
Tolerates new Infra Taint

iamkirkbater:
Adds a new toleration for the upcoming infra taint change.  I'm leaving the old toleration because from my testing if the new taint is not pushed out these pods will be pushed out to worker nodes, but if the old toleration is left it will still land on the infra nodes.

If this is not the correct place to make this change I'm more than happy to learn where the correct place is :)

This work is related to https://issues.redhat.com/browse/OSD-2927


cblecker:
/lgtm
/hold

/assign @mbarnes 
Can you confirm this won't conflict? If so, I'm fine with this going in.

PR #105, opened 2020-10-13
OSD-5408 implement ensureDNSForService for gcp

yithian:


pkg/cloudclient/gcp/private.go
```diff
@@ -48,7 +51,37 @@ func (c *Client) setDefaultAPIPublic(ctx context.Context, kclient client.Client,
...
 func (c *Client) ensureDNSForService(ctx context.Context, kclient client.Client, svc *corev1.Service, dnsName, dnsComment string) error {
-	return errors.New("ensureDNSForService is not implemented")
+	// google.golang.org/api/dns/v1.Service is a struct, not an interface, which
+	// will make this all but impossible to write unit tests for
+
+	dnsChange := &gdnsv1.Change{
+		Additions: []*gdnsv1.ResourceRecordSet{
+			{
+				Name:    dnsName,
+				Rrdatas: svc.Spec.ExternalIPs,
```
mbarnes, line 61: I think the resource record IPs need to come from the service's status, since the service creates the load balancer and the load balancer IPs are not known ahead of time.

Use `svc.Status.LoadBalancer.Ingress[n].IP` (will need to build a string slice).

mbarnes (changes requested):

mbarnes (approved):
LGTM (without testing)

PR #106, opened 2020-10-13
Remove csheremeta from OWNERS

csheremeta:
since I haven't actually been generally involved in this repo :)


lisa:
/approve
/lgtm

we'll have to get you involved and added back :)

PR #108, opened 2020-10-14
OSD-5408 remove dns for service

yithian:


mbarnes (approved):
LGTM!  Man, that is so much simpler than AWS (assuming it works :wink:).

Also, thank you for the separate commit for `getIPAddressesFromService`.

PR #109, opened 2020-10-14
remove unnecessary params from google cloud dns functions

yithian:


mbarnes (approved):
Huh, surprised GCP DNS doesn't allow for comments in the resource records.

/lgtm

PR #111, opened 2020-10-15
Simplify ensureIngressControllersExist

fahlmant:
Simplifies the logic of `ensureIngressControllersExist` to help find nil pointer issues.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -371,29 +371,44 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 
 // ensure ingresscontrollers on publishingstrategy CR are present on the cluster
 func (r *ReconcilePublishingStrategy) ensureIngressControllersExist(appIngressList []cloudingressv1alpha1.ApplicationIngress, ingressControllerList *operatorv1.IngressControllerList) bool {
-	isContained := true
-	for _, app := range appIngressList {
-		var exists bool
-		for _, ingress := range ingressControllerList.Items {
-			// prevent nil pointer error
-			if ingress.Spec.Domain == "" ||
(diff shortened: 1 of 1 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -371,29 +371,44 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-			capListening := strings.Title(strings.ToLower(listening))
-			if ingress.Spec.Domain == app.DNSName && capListening == string(ingress.Status.EndpointPublishingStrategy.LoadBalancer.Scope) {
-				exists = true
-			}
+
+	for _, appIngress := range appIngressList {
+		result := doesIngressControllerExist(appIngress, ingressControllerList)
+
+		if !result {
+			return false
```
lisa, line 379: Can this be 

```
if !doesIngressControllerExist(appIngress, ingressControllerList) {
  return false
}
```

Is there a reason to keep `result` around?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -371,29 +371,44 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+}
+
+func doesIngressControllerExist(appIngress cloudingressv1alpha1.ApplicationIngress, ingressControllerList *operatorv1.IngressControllerList) bool {
+
+	for _, ingress := range ingressControllerList.Items {
+
+		// prevent nil pointer error
+		if !validateIngress(ingress) {
+			return false
+		}
```
lisa, line 393: If we're validating an Ingress here what does it mean for one to be invalid? How might the `ingressControllerList` come to have an "invalid" Ingress in it?

I'd like to see a comment here covering those topics because I can read the prevent nil pointer error comment and understand that from the `validateIngress` function but that doesn't tell me why it's important and how an invalid Ingress could get this far.

PR #114, opened 2020-10-16
Update toleration for operator and catalog to deploy to infra

jewzaam:
https://issues.redhat.com/browse/OSD-5545

### DIFF
--- a/deploy/50_cloud-ingress-operator.Deployment.yaml
+++ b/deploy/50_cloud-ingress-operator.Deployment.yaml
@@ -23,7 +23,7 @@ spec:
                 operator: Exists
             weight: 1
       tolerations:
-      - operator: Equal
+      - operator: Exists
         key: node-role.kubernetes.io/infra
         effect: NoSchedule
       containers:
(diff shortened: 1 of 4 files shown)


jewzaam:
Rebased, @lisa ptal

lisa:
@jewzaam what's changed since my /lgtm a few hours ago?

deploy/60_cloud-ingress-operator-registry.CatalogSource.yaml
```diff
@@ -18,9 +18,8 @@ spec:
         weight: 1
   tolerations:
   - operator: Equal
```
cblecker, line 20: Will this work with equal if there is no value? Should this be exists?

deploy/50_cloud-ingress-operator.Deployment.yaml
```diff
@@ -24,12 +24,7 @@ spec:
             weight: 1
       tolerations:
       - operator: Equal
```
cblecker, line 26: Will this work with equal if there is no value? Should this be exists?

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -312,9 +312,8 @@ objects:
               weight: 1
         tolerations:
         - operator: Equal
```
cblecker, line 314: Will this work with equal if there is no value? Should this be exists?

PR #115, opened 2020-10-19
publishingstrategy: Improve IngressControllerList log output

mbarnes:
Serialize the object as JSON so the log reveals the contents of pointer struct fields, particularly `EndpointPublishingStrategy`.
Currently the log only shows an address for these fields.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -1,6 +1,7 @@
 package publishingstrategy
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -136,7 +139,9 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 	}
 
 	// ----------------------------TODO: remove these debug logs afterwards --------------------------------------
-	log.Info(fmt.Sprintf("initial ingresscontroller list: %+v", ingressControllerList.Items))
+	var serializedIngressControllerList bytes.Buffer
+	_ = serializer.Encode(ingressControllerList, &serializedIngressControllerList)
+	log.Info(fmt.Sprintf("initial ingresscontroller list: %s", serializedIngressControllerList.String()))
```
lisa, line 144: Does this fail safe should the encoding fail for some reason?

mbarnes: I don't expect that to happen since the object is fetched, but if it did then the `bytes.Buffer` would just be empty.

lisa (approved):
/approve
/lgtm

PR #118, opened 2020-10-20
Support Updating Existing IngressController

dustman9000:
This change depends on https://github.com/openshift/cluster-ingress-operator/pull/472 (targeted for 4.7) and https://github.com/openshift/cluster-ingress-operator/pull/482 (for 4.6).

This changes the way cloud-ingress-operator reconciles the IngressController CR and removes the need to delete all CRs before adding new CRs (which was highly disruptive).

Tested on a stage 4.6 cluster (drow-10-20-20) with both modified operators.

Update 4-13-21:
Changes to support mutable loadbalancer scope were pulled from the OCP 4.7 and 4.6 releases due to security concerns. Work has been restarted to address this in a more secure manner here: https://github.com/openshift/cluster-ingress-operator/pull/582

### DIFF
--- a/deploy/20_cloud-ingress-operator.ClusterRole.yaml
+++ b/deploy/20_cloud-ingress-operator.ClusterRole.yaml
@@ -26,6 +26,7 @@ rules:
     - watch
     - delete
     - create
+    - update
 - apiGroups:
   - config.openshift.io
   resources:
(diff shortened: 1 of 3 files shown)


cblecker:
For clarity: this cannot merge until the entire fleet has been upgraded to include this fix.

dustman9000:
Since the upstream dependencies were removed from the 4.6 and 4.7 OCP releases due to a security concern and the new PR around supporting changing loadbalancer scope requires admin intervention to delete the service, I'm going to close this PR in favor of a new PR to support https://github.com/openshift/cluster-ingress-operator/pull/582 later.

PR #119, opened 2020-10-21
publishingstrategy: Check Domain value in Status, not Spec

mbarnes:
The initial "default" IngressController has no Domain value in its Spec section, only its Status section.  This causes the controller
to unnecessarily delete and recreate the object.

See my findings in [OSD-5507](https://issues.redhat.com/browse/OSD-5507).


mbarnes:
Holding since this might not be the whole story.

The controller also looks for a `Spec.DefaultCertificate.Name`, which is also absent in the initial "**default**" IngressController and could even crash the controller since it's not checking if `Spec.DefaultCertificate` is a `nil` pointer.

mbarnes:
Holding since @cblecker said he would prefer a more comprehensive overhaul of the comparison logic, so I'll likely end up closing this PR.

lisa:
@mbarnes @cblecker What is the status of this PR?

cblecker:
This isn't needed anymore due to the more robust rewrite
/close

PR #123, opened 2020-11-02
Refactor PublishingStrategy to simplify logic

fahlmant:
Essentially a full refresh of the AppIngress logic. New workflow is:

For each definition in ApplicationIngress list:
- Check if it exists on cluster. If it doesn't create it
- If it does, check if the specs match
- If they don't, delete it, unless its default. If its default check the status
    - If the default status doesn't match the desired spec delete it

This makes the controller close to idempotent and simplifies the logic. The `defaultAPIServerIngress` logic in untouched.

### DIFF
--- a/deploy/20_cloud-ingress-operator.ClusterRole.yaml
+++ b/deploy/20_cloud-ingress-operator.ClusterRole.yaml
@@ -24,6 +24,7 @@ rules:
     - get
     - list
     - watch
+    - patch
     - delete
     - create
 - apiGroups:
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -1,20 +1,19 @@
 package publishingstrategy
 
 import (
-	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"strings"
-	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cloud-ingress-operator/pkg/apis/cloudingress/v1alpha1"
 	cloudingressv1alpha1 "github.com/openshift/cloud-ingress-operator/pkg/apis/cloudingress/v1alpha1"
 	"github.com/openshift/cloud-ingress-operator/pkg/cloudclient"
 	baseutils "github.com/openshift/cloud-ingress-operator/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -35,6 +34,11 @@ const (
 
 var log = logf.Log.WithName("controller_publishingstrategy")
 var serializer = json.NewSerializerWithOptions(nil, nil, nil, json.SerializerOptions{})
+
+type patchField string
+
+var IngressControllerSelector patchField = "IngressControllerSelector"
+var IngressControllerCertificate patchField = "IngressControllerCertificate"
 
 /**
 * USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
@@ -105,7 +109,7 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
-	// get a list of all ingress on the cluster
+	// Get all IngressControllers on cluster with an annotation that indicates cloud-ingress-operator owns it
 	ingressControllerList := &operatorv1.IngressControllerList{}
 	listOptions := []client.ListOption{
 		client.InNamespace("openshift-ingress-operator"),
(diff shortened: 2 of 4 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -214,297 +184,69 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-
-// checkExistingIngress returns false if applicationIngress do not match any existing ingresscontroller on cluster
-func checkExistingIngress(existingMap map[string]operatorv1.IngressController, publishingStrategyIngress *cloudingressv1alpha1.ApplicationIngress) bool {
-	if _, ok := existingMap[publishingStrategyIngress.DNSName]; !ok {
-		log.Info(fmt.Sprintf("IngressController for %q not found", publishingStrategyIngress.DNSName))
+	if !(desiredSpec.EndpointPublishingStrategy == ingressController.Status.EndpointPublishingStrategy) {
 		return false
 	}
-	if !isOnCluster(publishingStrategyIngress, existingMap[publishingStrategyIngress.DNSName]) {
+	if !(mapToString(desiredSpec.RouteSelector.MatchLabels) == ingressController.Status.Selector) {
```
mbarnes, line 239: Suggestion: Instead of converting to a string, you could also use [metav1.ParseToLabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ParseToLabelSelector) on `ingressController.Status.Selector` and compare them with [reflect.DeepEqual](https://golang.org/pkg/reflect/#DeepEqual).  Then you wouldn't need `mapToString`.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,80 +105,50 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-	var ingressNotOnCluster []cloudingressv1alpha1.ApplicationIngress
-
-	exisitingIngressMap := convertIngressControllerToMap(ingressControllerList.Items)
-
-	// loop through every applicationingress in publishing strategy
-	for _, publishingStrategyIngress := range instance.Spec.ApplicationIngress {
-		if !checkExistingIngress(exisitingIngressMap, &publishingStrategyIngress) {
-			ingressNotOnCluster = append(ingressNotOnCluster, publishingStrategyIngress)
+		// Set the namespaced name
+		namespacedName := types.NamespacedName{Name: getIngressName(ingressDefinition.DNSName), Namespace: ingressControllerNamespace}
```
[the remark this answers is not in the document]
fahlmant: :+1: to that. getIngressName was the only function I really didn't look at changing.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: Ah that's perfect. Thank ypu

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,80 +105,50 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		// If the IngressController exists, ensure it matches PublishingStrategies definition
+		if !reflect.DeepEqual(ingressController.Spec, desiredIngressController.Spec) {
+			// Deafult is a special case, the matching data can also be in the status
+			if ingressDefinition.Default {
+				if checkDefaultIngressControllerStatus(*ingressController, desiredIngressController.Spec) {
+					continue
+				}
+			}
+			// If the definitions aren't the same, delete the IngressController
+			err = r.client.Delete(context.TODO(), ingressController)
```
sam-nguyen7, line 144: we only want to delete if the dns and/or listening(internal/external) change

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -214,297 +184,69 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		return false
 	}
-	return ingressMap
-}
-
-// checkExistingIngress returns false if applicationIngress do not match any existing ingresscontroller on cluster
-func checkExistingIngress(existingMap map[string]operatorv1.IngressController, publishingStrategyIngress *cloudingressv1alpha1.ApplicationIngress) bool {
-	if _, ok := existingMap[publishingStrategyIngress.DNSName]; !ok {
-		log.Info(fmt.Sprintf("IngressController for %q not found", publishingStrategyIngress.DNSName))
+	if !(desiredSpec.EndpointPublishingStrategy == ingressController.Status.EndpointPublishingStrategy) {
```
sam-nguyen7, line 236: can we assume `ingressController.Status.EndpointPublishingStrategy` always exist?

sam-nguyen7: I'd also check for capitalization here (ie internal vs Internal)

fahlmant: This is only checked for the default ingresscontroller if the info is not in the spec. AFAIK the default IngressController CR will always have it in one or the other.

fahlmant: Also, this is checking the generated IngresController, so both of these are the same types, either `ExternalLoadBalancer` or `InternalLoadBalancer` not the appInress directly.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: Is there a desgin doc for this? OSD-5507 mentions using DeepEqual

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: ah ok so its when we are checking the spec that we would need to worry about capitalization then :+1:

pkg/controller/publishingstrategy/publishingstrategy_controller.go
sam-nguyen7: I made a comment on the card I believe. No design doc, but we shouldn't delete if we can patch imo because delete leads to dns blips which is nice to avoid

cblecker: +1 to Sam. We should only delete if it's an immutable field (domain and endpointpublishingstrategy I think are the only two).

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,80 +105,50 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+				// If the CR was created, requeue PublishingStrategy
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, err
 		}
 
-		if appingress.Default == true {
-			err := r.defaultIngressHandle(appingress, ingressControllerList, newCertificate)
+		// If the IngressController exists, ensure it matches PublishingStrategies definition
+		if !reflect.DeepEqual(ingressController.Spec, desiredIngressController.Spec) {
```
cblecker, line 136: There are certain fields like `Spec.Domain` that the installer leaves blank, which allows them to be the default (you can see what the default is in `Status.Domain`.

The other thing we have to do if we are doing this is enumerate *all* the spec fields to ensure we are setting them appropriately. There are some fields we currently don't manage/compare with the existing logic, and a shift to using DeepEqual, while ultimately will fix this once and for all, could break functionality on the non-compared fields if we clobber them.

fahlmant: @cblecker The default CR gets a spec check in this workflow currently, whic covers the first part of what you're talking about.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,78 +109,87 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-	// since we don't own ingresscontroller object, adding this check ensures that the proper ingresscontrollers have been created by cluster-ingress-operator
-	// if check fails, requeue the reconcile and try to re-create the ingresscontrollers until successful
-	if !r.ensureIngressControllersExist(instance.Spec.ApplicationIngress, ingressList) {
-		reqLogger.Info("IngressController does not match PublishingStrategy. Requeue and try again")
-		return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		// Check Spec fields that CAN be patched vs desired IngressController
+		if valid, field := validatePatchableSpec(*ingressController, desiredIngressController.Spec); !valid {
+			baseToPatch := client.MergeFrom(ingressController.DeepCopy())
+			if ingressDefinition.Default {
+				if valid, field := validatePatchableStatus(*ingressController, desiredIngressController.Spec); valid {
```
csheremeta, line 166:
```suggestion
				if valid, field := validatePatchableStatus(*ingressController, desiredIngressController.Spec); !valid {
```

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,78 +109,87 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
 
-	// since we don't own ingresscontroller object, adding this check ensures that the proper ingresscontrollers have been created by cluster-ingress-operator
-	// if check fails, requeue the reconcile and try to re-create the ingresscontrollers until successful
-	if !r.ensureIngressControllersExist(instance.Spec.ApplicationIngress, ingressList) {
-		reqLogger.Info("IngressController does not match PublishingStrategy. Requeue and try again")
-		return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		// Check Spec fields that CAN be patched vs desired IngressController
+		if valid, field := validatePatchableSpec(*ingressController, desiredIngressController.Spec); !valid {
+			baseToPatch := client.MergeFrom(ingressController.DeepCopy())
+			if ingressDefinition.Default {
```
lisa, line 165: Out of curiosity why is there different behaviour between default ingress vs additional ones?

lisa (changes requested):
/hold

A bunch of requests around comments.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,78 +109,87 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-	_ = serializer.Encode(ingressControllerList, &serializedIngressControllerList)
-	log.Info(fmt.Sprintf("initial ingresscontroller list: %s", serializedIngressControllerList.String()))
-	log.Info(fmt.Sprintf("appingress on publishingstrategy CR: %+v", instance.Spec.ApplicationIngress))
-	log.Info(fmt.Sprintf("ingress to reconcile: %+v", ingressNotOnCluster))
-	// -----------------------------------------------------------------------------------------------------------
-
-	for _, appingress := range ingressNotOnCluster {
-		newCertificate := &corev1.LocalObjectReference{
-			Name: appingress.Certificate.Name,
+		// Let an IngressController with a deletion timestamp delete
```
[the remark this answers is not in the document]
fahlmant: I didn't add this line so no idea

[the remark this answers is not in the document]
fahlmant: cc @mbarnes @sam-nguyen7

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: "defult" CR also needs a status check because on initial creation the spec isn't filled out

pkg/controller/publishingstrategy/publishingstrategy_controller.go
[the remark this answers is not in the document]
mbarnes: I added that yesterday during the incident.  We were informed that ingress controllers can take awhile to delete while taking down cloud resources, so if it has a deletion timestamp then disregard it.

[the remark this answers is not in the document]
sam-nguyen7: when an ingresscontroller is being deleted, it takes time as it needs to delete several services (ie the load balancer service has finalizers for the cloud provider resource cleanup). If ingresscontroller still has these finalizers, there is no point in trying to re-create it. This check ensures ingresscontroller is deleted before we try and create so we don't fill up cluster-ingress-operator logs with a bunch of error messages

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,78 +109,122 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+		// Generate the desired IngressController spec based on the ApplicationIngress definition.
+		// This generated spec will be compared against the actual spec as desrcibed above
+		desiredIngressController := generateIngressController(ingressDefinition)
 
-	// loop through every applicationingress in publishing strategy
-	for _, publishingStrategyIngress := range instance.Spec.ApplicationIngress {
-		if !checkExistingIngress(exisitingIngressMap, &publishingStrategyIngress) {
-			ingressNotOnCluster = append(ingressNotOnCluster, publishingStrategyIngress)
+		// Attempt to find the IngressController referenced by the ApplicationIngress
+		// by doing a GET of the namespaced name object build above agains the k8s api.
```
lisa, line 150:
```suggestion
		// by doing a GET of the namespaced name object build above against the k8s api.
```

typo

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -105,78 +109,122 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-	// if check fails, requeue the reconcile and try to re-create the ingresscontrollers until successful
-	if !r.ensureIngressControllersExist(instance.Spec.ApplicationIngress, ingressList) {
-		reqLogger.Info("IngressController does not match PublishingStrategy. Requeue and try again")
-		return reconcile.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		// All the remaining fields are mutable and don't require a deletion of the IngresscController
+		// If any of the fields are differet, that field will be patched
+		if valid, field := validatePatchableSpec(*ingressController, desiredIngressController.Spec); !valid {
+			baseToPatch := client.MergeFrom(ingressController.DeepCopy())
+			if ingressDefinition.Default {
+				if valid, field := validatePatchableStatus(*ingressController, desiredIngressController.Spec); !valid {
```
lisa, line 201: Why isn't this function used for both? Is it wise to re-scope `field`?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -214,297 +262,96 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
-	return ingressMap
 }
 
-// checkExistingIngress returns false if applicationIngress do not match any existing ingresscontroller on cluster
-func checkExistingIngress(existingMap map[string]operatorv1.IngressController, publishingStrategyIngress *cloudingressv1alpha1.ApplicationIngress) bool {
-	if _, ok := existingMap[publishingStrategyIngress.DNSName]; !ok {
-		log.Info(fmt.Sprintf("IngressController for %q not found", publishingStrategyIngress.DNSName))
+func validateStaticStatus(ingressController operatorv1.IngressController, desiredSpec operatorv1.IngressControllerSpec) bool {
+
+	if !(desiredSpec.Domain == ingressController.Status.Domain) {
```
lisa, line 311:
```suggestion
	if desiredSpec.Domain != ingressController.Status.Domain {
```

Any reason not to do this?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -214,297 +262,96 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
 		return false
 	}
-	if !isOnCluster(publishingStrategyIngress, existingMap[publishingStrategyIngress.DNSName]) {
+	if !(desiredSpec.EndpointPublishingStrategy.LoadBalancer.Scope == ingressController.Status.EndpointPublishingStrategy.LoadBalancer.Scope) {
```
lisa, line 314:
```suggestion
	if desiredSpec.EndpointPublishingStrategy.LoadBalancer.Scope != ingressController.Status.EndpointPublishingStrategy.LoadBalancer.Scope {
```

Same here, and in the other functions.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: Used for both what?

lisa: For both "default" and "non-default." I suppose I don't understand the purpose of `validatePatchableSpec`.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: Both work and are understandable the same.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: @lisa As explained in the comments

The default IngressController is special. When its first created, the spec is not filled out.
Instead, the relavent fields are set in the status. For each of the spec checks, the status
will also be checked if the ApplicationIngress references the default IngressController

lisa: @fahlmant I think I was unclear in phrasing my question. What does `validatePatchableStatus` do? The function has no comments. What are the semantic meanings behind its two return values? What inputs cause certain kinds of outputs? Can it be used for both default and non-default ingress types?

It appears that there's duplication adjacent to this comment and if the `validatePatchableStatus` function can service both default and non-default ingresses that some of the duplication can be removed and thus the readability improved somewhat.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -214,297 +272,124 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+
+	return true
+}
+
+/* Compares the patchable desired Spec against the existing IngressController's status.
+The only patchable field that the ApplicationIngress controlls thats also stored in the IngressController's
+status is the RouteSelector. The default IngressController does not have those spec fields
+filled in at all when fisrt created. Instead, the default CRs status holds the correct values. This function
+is meant to be used when the default CRs spec is not filled in to see if the desired configuration is present in
+its status instead. Returns false if the RouteSelector in the status is empty or does not match the desired Spec.
```
lisa, line 364: When does it return true? When the route selector is set or matches the desired spec? What is second field in the return value?

fahlmant: The function returns a boolean, and is called validate. True returns when the status is valid, ie matches the desired spec.

deploy/20_cloud-ingress-operator.ClusterRole.yaml
```diff
@@ -24,6 +24,7 @@ rules:
     - get
     - list
     - watch
+    - patch
     - delete
     - create
```
cblecker, line 29: Should we simplify this to `- "*"`?

fahlmant: Happy to do it. Is there any specific benefit?

csheremeta: no, it's just easier for humans to read :)

sam-nguyen7 (commented):
lgtm. I was able to test manually and verify a lot of these cases. One case that specifically did not work for me was the deletion of ingress controller when deleting appIngress from PublishingStrategy CR. When removing apps2 from PS CR, the apps2 ingresscontroller on cluster did not get deleted, but it needs to. One thing to keep in mind, we only want to delete the ingresses that cloud-ingress-operator "owns" (ie. if another operator creates apps3 ingress, we want to leave that ingress alone).

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -430,3 +479,19 @@ func validatePatchableSpec(ingressController operatorv1.IngressController, desir
 
 	return true, ""
 }
+
+// Given an IngressControllerList, returns only IngressControllers with the Owner annotation
```
csheremeta, line 483:
```suggestion
// Given an IngressControllerList, returns only IngressControllers with the cloud-ingress-operator Owner annotation
```

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -430,3 +479,19 @@ func validatePatchableSpec(ingressController operatorv1.IngressController, desir
...
+func getIngressWithOwnerAnnotation(ingressList operatorv1.IngressControllerList) *operatorv1.IngressControllerList {
```
csheremeta, line 484: Are the IngressControllers with the Owner annotation and the IngressControllers with the Owner annotation set to CIO the same set? I want to assume they aren't, in which case I think this function may be better named something like `getIngressWithCIOOwnerAnnotation` - what are your thoughts?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -109,6 +109,36 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
...
+	match the list of ApplicationIngresses in the PublishingStrategy, a map is created to
+	tie IngressControllers existing on cluster with the owner annotation
+	to the ApplicationIngresses existing in the PublishingStrategy CR. If an IngressController CR
+	owned by cloud-ingress-operator exists on cluster but an associated ApplicationIngress does not exist in
+	the PublishingStrategy CR (likely removed from the list of ApplicationIngress),
+	that IngressController CR should be deleted.
+	As each ApplicationIngress is processed, and the corresponding IngressController CR is confirmed to exist,
+	that entry in the map will be turned `true`. When all ApplicationIngresses have been processed, any remaining
+	IngressController CRs in the map with the value `false` will be deleted.
+	*/
```
csheremeta, line 135: this is lovely :)

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -246,6 +279,22 @@ func (r *ReconcilePublishingStrategy) Reconcile(request reconcile.Request) (reco
 		}
 	}
 
+	// Delete all IngressControllers that are owned by cloud-ingress-operator but not in PublishingStrategy
+	for ingress, exists := range ownedIngressExistingMap {
+		if !exists {
```
csheremeta, line 284:
```suggestion
	for ingress, inPublishingStrategy := range ownedIngressExistingMap {
		if !inPublishingStrategy {
```
I think we could rename this variable to make the criteria for this `if` block more clear.

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: +1 to making it more specific. I don't think cluster ingres operator adds the annotations but in theory anyone could

pkg/controller/publishingstrategy/publishingstrategy_controller.go
fahlmant: +1

sam-nguyen7:
Tested this PR with @fahlmant on a cluster and everything is working as expected.

lisa:
> Tested this PR with @fahlmant on a cluster and everything is working as expected.

/lgtm

PR #124, opened 2020-11-03
Add GCP cloud client (DNS methods only)

cblecker:
This is just to get some tests running. It's a rebase of the cloudclient-gcp branch

**NOTE:** The `SetDefaultAPIPublic` and `SetDefaultAPIPrivate` methods for GCP are not part of this PR.  @blrm is working on this part and will submit a separate PR when ready.

### DIFF
--- a/deploy/10_cloud-ingress-operator-credentials-aws.CredentialsRequest.yaml
+++ b/deploy/10_cloud-ingress-operator-credentials-aws.CredentialsRequest.yaml
@@ -1,11 +1,11 @@
 apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
-  name: cloud-ingress-operator-credentials
+  name: cloud-ingress-operator-credentials-aws
   namespace: openshift-cloud-ingress-operator
 spec:
   secretRef:
-    name: cloud-ingress-operator-credentials
+    name: cloud-ingress-operator-credentials-aws
     namespace: openshift-cloud-ingress-operator
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1
--- a/deploy/10_cloud-ingress-operator-credentials-gcp.CredentialsRequest.yaml
+++ b/deploy/10_cloud-ingress-operator-credentials-gcp.CredentialsRequest.yaml
@@ -1,0 +1,15 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  name: cloud-ingress-operator-credentials-gcp
+  namespace: openshift-cloud-ingress-operator
+spec:
+  secretRef:
+    name: cloud-ingress-operator-credentials-gcp
+    namespace: openshift-cloud-ingress-operator
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/dns.admin
+    skipServiceCheck: true
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -41,11 +41,11 @@ objects:
     - apiVersion: cloudcredential.openshift.io/v1
       kind: CredentialsRequest
       metadata:
-        name: cloud-ingress-operator-credentials
+        name: cloud-ingress-operator-credentials-aws
         namespace: openshift-cloud-ingress-operator
       spec:
         secretRef:
-          name: cloud-ingress-operator-credentials
+          name: cloud-ingress-operator-credentials-aws
           namespace: openshift-cloud-ingress-operator
         providerSpec:
           apiVersion: cloudcredential.openshift.io/v1
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -86,6 +86,21 @@ objects:
             - route53:ListHostedZonesByName
             - route53:ListResourceRecordSets
             - route53:UpdateHostedZoneComment
+    - apiVersion: cloudcredential.openshift.io/v1
+      kind: CredentialsRequest
+      metadata:
+        name: cloud-ingress-operator-credentials-gcp
+        namespace: openshift-cloud-ingress-operator
+      spec:
+        secretRef:
+          name: cloud-ingress-operator-credentials-gcp
+          namespace: openshift-cloud-ingress-operator
+        providerSpec:
+          apiVersion: cloudcredential.openshift.io/v1
+          kind: GCPProviderSpec
+          predefinedRoles:
+          - roles/dns.admin
+          skipServiceCheck: true
     - apiVersion: rbac.authorization.k8s.io/v1
       kind: ClusterRole
       metadata:
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -134,6 +134,7 @@ objects:
         resources:
         - infrastructures
         - apiservers
+        - dnses
         verbs:
         - list
         - get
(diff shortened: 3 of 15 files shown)


pkg/errors/errors.go
```diff
@@ -10,9 +10,9 @@ type LoadBalancerNotFoundError struct {
 
 func (e *LoadBalancerNotFoundError) Error() string { return e.e }
 
-func NewLoadBalancerNotFoundError(lb string) error {
+func NewLoadBalancerNotFoundError() error {
 	return &LoadBalancerNotFoundError{
-		e: fmt.Sprintf("Could not find LoadBalancer %s", lb),
+		e: "Could not find LoadBalancer",
```
lisa, line 15: Don't we want to include the name of the LB we can't find?

lisa (changes requested):
Minor requests + questions.

I'll let other folks speak to the GCP bits.

pkg/cloudclient/gcp/private.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+
+	// The resource record set to search for / add.
+	newRRSet := &gdnsv1.ResourceRecordSet{
+		Name:    dnsName + "." + baseDomain + ".",
+		Rrdatas: svcIPs,
+		Type:    "A",
+		Ttl:     30,
```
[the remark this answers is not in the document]
mbarnes: Yes, TTL is in seconds.  I based this on [cluster-ingress-operator's default TTL value](https://github.com/openshift/cluster-ingress-operator/blob/3fbc377049ee88674e09da6ab111d134467e1392/pkg/operator/controller/ingress/dns.go#L22-L28).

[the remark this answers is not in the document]
lisa: Oh wow, okay. That's shorter than I would have expected. 👍

pkg/errors/errors.go
mbarnes: The GCP methods never learn the name of the load balancer.  They just take the IPs from the Service object passed in.

lisa: @mbarnes I'd support that change. Is that something you'd like to make in this PR, or separately?

mbarnes: Sure, I'll add it here.

mbarnes:
Force-pushed with the above discussed changes added to existing commits (to keep the history clean).

deploy/10_cloud-ingress-operator-credentials-gcp.CredentialsRequest.yaml
```diff
@@ -0,0 +1,18 @@
...
+  secretRef:
+    name: cloud-ingress-operator-credentials-gcp
+    namespace: openshift-cloud-ingress-operator
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/iam.serviceAccountUser
+    - roles/compute.loadBalancerAdmin
+    - roles/compute.securityAdmin
```
cblecker, line 16: I don't think we need these
```suggestion
```

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -119,6 +137,7 @@ objects:
         resources:
         - infrastructures
         - apiservers
+        - dnses
```
cblecker, line 140: Is this required?

pkg/cloudclient/gcp/private.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if len(ips) == 0 {
+		return nil, cioerrors.NewLoadBalancerNotReadyError()
+	}
+
+	return ips, nil
+}
+
+func equalResourceRecordSets(rrset1, rrset2 *gdnsv1.ResourceRecordSet) bool {
+	// Just compare the fields we set ourselves.
```
cblecker, line 193: For clarity, maybe add a comment that if any of these expressions are false, then you're returning false. We only return true if every expression is true.

hack/olm-registry/olm-artifacts-template.yaml
mbarnes: That part is, yes.  I added that after getting a permission error.

deploy/10_cloud-ingress-operator-credentials-gcp.CredentialsRequest.yaml
mbarnes: Ack, they were just blind guesses.  I'll test without them and verify.

pkg/cloudclient/gcp/private.go
mbarnes: Agreed.  I'd like to play with it a bit more and see if `reflect.DeepEqual` could work here.  I haven't yet observed GCP altering or adding info in these structs.

deploy/10_cloud-ingress-operator-credentials-aws.CredentialsRequest.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
-  name: cloud-ingress-operator-credentials
+  name: cloud-ingress-operator-credentials-aws
   namespace: openshift-cloud-ingress-operator
 spec:
   secretRef:
-    name: cloud-ingress-operator-credentials
+    name: cloud-ingress-operator-credentials-aws
```
lisa, line 8: How do we expect this to work in terms of migrating to a new version? Presumably we'll have an old `CredentialsRequest`/`Secret` hanging around (named `cloud-ingress-operator-credentials`) in the AWS clusters? Do we need to do cleanup in this PR or otherwise?

mbarnes: I was thinking we clean it up in [managed-cluster-config](https://github.com/openshift/managed-cluster-config).  I'd prefer that kind of (temporary?) cruft go there instead of polluting operator code.

deploy/10_cloud-ingress-operator-credentials-gcp.CredentialsRequest.yaml
mbarnes: Looks like you're right.

pkg/cloudclient/gcp/private.go
mbarnes: Got `reflect.DeepEqual` working on the record comparisons.

mbarnes:
Latest force-push rejiggers (technical term) the DNS logic to make `reflect.DeepEqual` work for record comparisons and also avoid unnecessary Google DNS calls, as revealed by additional log messages.

pkg/cloudclient/gcp/private.go
```diff
@@ -0,0 +1,198 @@
...
+		// An empty slice will proceed directly to Create.
+		for _, rrset := range response.Rrsets {
+			if reflect.DeepEqual(newRRSet, rrset) {
+				dnsChange.Additions = []*gdnsv1.ResourceRecordSet{}
+			} else {
+				dnsChange.Deletions = append(dnsChange.Deletions, rrset)
+			}
+		}
+
+		if len(dnsChange.Additions) > 0 {
```
cblecker, line 123: Does this need to change if there are deletions?

mbarnes: No, because if there's any deletions then there's guaranteed to be an addition.

mbarnes:
Force-pushed again to fix a bug I found in `removeDNSForService` causing DNS cleanup to be mistakenly skipped.
I think I'm done messing with things, sorry for the churn during a review.

deploy/10_cloud-ingress-operator-credentials-aws.CredentialsRequest.yaml
mbarnes: Opened [OSD-5792](https://issues.redhat.com/browse/OSD-5792) for this part.

cblecker:
Oh right.. I'm technically the "author" of the PR, although @mbarnes has been doing the work on it. Manually adding the label..

PR #126, opened 2020-11-19
GCP methods for setDefaultAPIPublic and setDefaultAPIPrivate for cloudclient interface

blrm:
This PR implements the `setDefaultAPIPublic` and `setDefaultAPIPrivate` methods for GCP in the cloudclient interface. There are a few notable differences from how this works in AWS:

* GCP network load balancers don't have automatically generated A records assigned to them. Instead, they just have a public IP associated with them. So instead of updating an alias record, this just updates the existing A record for the API endpoint and points it to the relevant load balancer's IP. 
* The GCP API treats a load balancer as a combination of two resources: a frontend (ForwardingRule), and a backend (TargetPool or BackendService). A ForwardingRule that specifies a TargetPool or BackendService represents a load balancer. `setDefaultAPIPrivate` deletes the ForwardingRule and releases the static IP associated with it when switching to a private cluster, but leaves the TargetPool in place. `setDefaultAPIPublic` reserves a new static IP and creates a ForwardingRule that targets the pre-existing TargetPool.


pkg/cloudclient/gcp/private.go
```diff
@@ -201,3 +283,245 @@ func getIPAddressesFromService(svc *corev1.Service) ([]string, error) {
...
+	}
+	return intIPAddress, nil
+}
+
+func getClusterRegion(kclient client.Client) (string, error) {
+	infra, err := baseutils.GetInfrastructureObject(kclient)
+	if err != nil {
+		return "", err
+	}
+	return infra.Status.PlatformStatus.GCP.Region, nil
```
mbarnes, line 335: This can be addressed later, but the base `utils` package also has a `GetClusterRegion` function.  It currently assumes AWS, so it needs extended for GCP with a "switch" statement on the platform type, but would be preferable to use that instead.

pkg/cloudclient/gcp/private.go
```diff
@@ -201,3 +283,245 @@ func getIPAddressesFromService(svc *corev1.Service) ([]string, error) {
...
+	return nil
+}
+
+func (c *Client) updateAPIARecord(kclient client.Client, recordName string, newIP string) (oldIP string, err error) {
+	clusterDNS := &configv1.DNS{}
+	err = kclient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, clusterDNS)
+	if err != nil {
+		return "", err
+	}
+	pubZoneRecords, err := c.dnsService.ResourceRecordSets.List(c.projectID, clusterDNS.Spec.PublicZone.ID).Do()
```
mbarnes, line 496: Suggestion: If I'm understanding this right, I think you could simplify the logic here with:
```
c.dnsService.ResourceRecordSets.List(...).Name(recordName).Do()
```
That would limit the result set to records matching `recordName` so you don't have to hunt for them further down.

mbarnes (commented):
Couple cleanup suggestions, but nothing that should block merging.

Not real familiar with all the GCP mechanics involved, but the code looks good to me fwiw.

cblecker:
Nothing blocking from me either.

/lgtm
/approve

PR #128, opened 2020-12-11
Check ForwardingRule names to support multiple clusters in a single account

blrm:
Testing on GCP in an account with multiple installed clusters showed issues with identifying the correct load balancer forwardingrule when toggling the API endpoint between public and private. This PR just adds a conditional that confirms the name of the forwardingrule to make sure it is a rule that is associated with the intended cluster.

### DIFF
--- a/pkg/cloudclient/gcp/private.go
+++ b/pkg/cloudclient/gcp/private.go
@@ -100,15 +100,6 @@ func (c *Client) setDefaultAPIPublic(ctx context.Context, kclient client.Client,
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


lisa:
/assign @mbarnes 
/assign @yithian 

It's probably better to have the co-authors of the file review first.

lisa:
/cc @yithian 
/cc @mbarnes 

Sorry for the spam, I figured the bot would pick you both up as reviewers with the /assign.

cblecker:
I wish we had the option of labels or something rather than relying on the name, but gcp doesn't offer that.

/lgtm
/approve

PR #129, opened 2020-12-14
Add safety check to protect against malformed publishingstrategy

cblecker:
This adds an additional check to default ingress definitions to protect against a malformed publishing strategy being added to a cluster.


fahlmant (approved):
/lgtm

PR #134, opened 2021-01-20
aws: Change listAllNLBs to listOwnedNLBs

mbarnes:
Only return load balancers owned by the cluster, according to tags.
CCS clusters may have customer-owned load balancers that should be left alone.

Also, assume there may be a large number of load balancers present, or at least enough to require multiple `DescribeLoadBalancers` and `DescribeTags` calls.

Fixes [OSD-6242](https://issues.redhat.com/browse/OSD-6242)


mbarnes:
Hold while verifying this on a staging cluster.

fahlmant:
Looks good at a first glance

mbarnes:
I manually verified the change in the staging environment with the following procedure:

- Create an internet-facing "decoy" NLB in the same VPC as the cluster.
- In OCM, select "Make API private" and apply the change

The current **cloud-ingress-operator** deletes both the cluster's external master API load balancer and my "decoy" that the cluster does not own.  The patched **cloud-ingress-operator** only deletes the cluster's external master API load balancer; the "decoy" is untouched.

Removing **hold** label.

lisa (approved):
/lgtm
/approve

PR #140, opened 2021-02-18
Dockerfile fix

fahlmant:
registry.ci is the new image registry that replaces (old)registry.svc as per https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/. We can transition to use the new as the public images are supported and available by CI.


lisa:
/lgtm
/approve

i look forward to the bot to stop spamming me every hour!

PR #141, opened 2021-02-22
RBAC permissions for sre-sshd client auth

gsleeman:
These permissions are required before sre-ssh-proxy-container can be used for client auth

https://github.com/openshift/sre-ssh-proxy-container/pull/24/commits/022519eaf0e90a62aee3ef8d49e3306cd4bbd0e6

### DIFF
--- a/deploy/35_sre-sshd-signer.Role.yaml
+++ b/deploy/35_sre-sshd-signer.Role.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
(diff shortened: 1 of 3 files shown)


mwoodson:
is there a jira ticket to reference with this? I'm unaware of what is trying to be done here.

gsleeman:
There's no jira issue for this, though it relates to OSD-6534

The rbac here is to allow the ssh container to issue certificate authenticating users to the apiserver, it's intended as a secondary means of cluster authentication in the case of an ocm outage

There's a few parts involved including this change in CIO

PR #144, opened 2021-03-08
Metrics to record APISchemes

sedroche:
Jira: https://issues.redhat.com/browse/OSD-6591

backplane relies on the rh-api endpoint to communicate with target clusters. There has [been a few times](https://issues.redhat.com/browse/OSD-6540) where we discovered the rh-api APIScheme was missing.
This PR will add metrics to record what APISchemes are in the cluster.
backplane team will add an alert for rh-api specifically based on this metric

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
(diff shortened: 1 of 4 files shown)


lisa:
@sedroche How does this implementation compare to those available in [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics/blob/8e35f790041384c12f36b40c47bae05e15c6f243/docs/validatingwebhookconfiguration-metrics.md)?

PR #145, opened 2021-03-15
Update controller to apply infra nodeSelectors to ingresscontrollers

blrm:
For https://issues.redhat.com/browse/OSD-5928, this updates the operator to patch the correct nodeSelector and tolerations onto any IngressControllers that it manages. This nodeSelector and toleration will ensure that all router pods are scheduled onto infra nodes. 

The unit tests have also been updated to account for the change in fields being managed on ingresscontrollers.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -30,6 +30,7 @@ import (
 const (
 	defaultIngressName         = "default"
 	ingressControllerNamespace = "openshift-ingress-operator"
+	infraNodeLabelKey          = "node-role.kubernetes.io/infra"
 )
 
 var log = logf.Log.WithName("controller_publishingstrategy")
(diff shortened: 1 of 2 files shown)


pkg/controller/publishingstrategy/publishingstrategy_controller_test.go
```diff
@@ -421,6 +421,18 @@ func TestValidatePatchableSpec(t *testing.T) {
 			DefaultCertificate: &corev1.LocalObjectReference{
 				Name: "example-cert-nondefault",
 			},
+			NodePlacement: &operatorv1.NodePlacement{
+				NodeSelector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"node-role.kubernetes.io/infra": ""},
```
dustman9000, line 426: Since we're using this string twice, can we turn it into a const?

pkg/controller/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -371,6 +376,18 @@ func generateIngressController(appIngress v1alpha1.ApplicationIngress) *operator
 			DefaultCertificate: &corev1.LocalObjectReference{
 				Name: appIngress.Certificate.Name,
 			},
+			NodePlacement: &operatorv1.NodePlacement{
+				NodeSelector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"node-role.kubernetes.io/infra": ""},
```
dustman9000, line 381: Can we turn "node-role.kubernetes.io/infra" into a const?

dustman9000:
Looks good! Just a minor comment that shouldn't block this PR.
/lgtm

PR #149, opened 2021-03-15
Fix golangci-lint Errors

dustman9000:
This fixes the more pedantic golangci-lint errors that will be added in the updated prow config: https://github.com/openshift/release/pull/16830


2uasimojo:
Code changes are legit. Tested locally, passes.

/lgtm

PR #152, opened 2021-03-17
Regenerate openapi with absolute paths

2uasimojo:
We're still trying to figure out why `make openapi-generate` sometimes injects absolute paths (which is what CI does, and therefore expects) and sometimes relative. Meanwhile, this should make the `validate` CI check pass.


dustman9000:
Thanks Eric!
/lgtm

cblecker:
actually
/hold

your git author is messed up @2uasimojo

PR #155, opened 2021-03-22
Rely on instance ID rather than subnet ID

fahlmant:
https://issues.redhat.com/browse/OSD-6790

Currently, the flow of the code is
1. Get master machine CR
2. Get Subnet.ID from the provider spec
3. Use the Subnet ID to call DescribeSubnets
4. Use the first element of the output from 3 to get the VPC ID
5. Use the VPC ID to list all subnets and route tables within it
6. Check each subnet to see if its "public"

In the master machine CR, Subnet.ID shows up for BYO VPC clusters, but not for installer-created VPC clusters (instead it shows Subnet.Tag with the name of the subnet). What happens if Subnet.ID is missing is that in step 3, DescribeSubnets returns ALL subnets. This is ok in scenarios where there's only one VPC in the region because all subnets are guaranteed to be within the VPC. If there's more than one VPC though, there's no guarantee what VPC the first subnet will be a part of. 

To remedy this, this PR moves to use the instance ID of the master machine directly, then calls DescribeInstance on that instance which has the VPC ID in the response.

### DIFF
--- a/pkg/cloudclient/aws/private.go
+++ b/pkg/cloudclient/aws/private.go
@@ -329,29 +329,32 @@ func (c *Client) getPublicSubnets(kclient client.Client) ([]string, error) {
 	// Get the first master machine in the list
 	masterMachine := machineList.Items[0]
 
-	// Get the AWS ProviderSpec from the machine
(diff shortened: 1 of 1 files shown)


mwoodson (approved):
/lgtm

PR #160, opened 2021-04-09
[WIP] Switch to boilerplate for container and catalog build

jharrington22:
Do not merge

Contains a second commit with fixes that be resolved in upcoming boilerplate releases

The rest I've attempted to break up into logical changes. This deploys as is and OLM is happy but I need to clean up what is left in `deploy/` and ensure the right resources are applied on cluster.


Makefile
```diff
@@ -3,13 +3,13 @@ include boilerplate/generated-includes.mk
...
-CONTAINER_ENGINE?=docker
+CONTAINER_ENGINE=docker
 
 # Needed for build-catalog-image
 CATALOG_REGISTRY_ORGANIZATION?=app-sre
 
-# TODO: remove this include once boilerplate supports generating package
-include functions.mk
+# Enable boilerplate OLM CSV generate script
+DEFAULT_CSV_GENRATOR=common
```
2uasimojo, line 12: This isn't a thing.

```
[efried@efried boilerplate]$ git grep DEFAULT_CSV_GEN
[efried@efried boilerplate]$ 
```

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
```diff
@@ -18,19 +18,36 @@ while getopts "o:H:c:" option; do
...
 # Checking parameters
-check_mandatory_params operator_channel operator_name  
+check_mandatory_params operator_channel operator_name
 
 # Parameters for the Dockerfile
 SAAS_OPERATOR_DIR="saas-${operator_name}-bundle"
+BUNDLE_DIR="${SAAS_OPERATOR_DIR}/${operator_name}"
 REGISTRY_IMG="quay.io/app-sre/${operator_name}-registry"
 DOCKERFILE_REGISTRY="Dockerfile.olm-registry"
+OPERATOR_NEW_VERSION=$(ls "${BUNDLE_DIR}" | sort -t . -k 3 -g | tail -n 1)
```
2uasimojo, line 28: This should go after the dir check below

boilerplate/openshift/golang-osd-operator/csv-generate/csv-generate.mk
```diff
@@ -2,22 +2,22 @@
 	
 .PHONY: staging-hack-csv-build
 staging-hack-csv-build: 
-	@${CONVENTION_DIR}/csv-generate/csv-generate.sh -o $(OPERATOR_NAME) -i $(OPERATOR_IMAGE) -c staging -H $(CURRENT_COMMIT) -n $(COMMIT_NUMBER) -g hack
+	@${CONVENTION_DIR}/csv-generate/csv-generate.sh -o $(OPERATOR_NAME) -i $(OPERATOR_IMAGE_URI) -c staging -H $(CURRENT_COMMIT) -n $(COMMIT_NUMBER) -g hack
```
2uasimojo, line 5: In https://github.com/openshift/boilerplate/pull/134 I reverted this and added a param for `-V $(OPERATOR_VERSION)` to facilitate computing the by-digest URI -- see https://github.com/openshift/boilerplate/pull/134/files#diff-424ddf34b68cef0633f4520f756a919d70a24880b5ab931d3cd85d17f8659b5fR38-R58

config/templates/csv-template.yaml
```diff
@@ -0,0 +1 @@
+../../hack/templates/csv.yaml
```
2uasimojo, line 1: Meh, why not just move it?

hack/generate-operator-bundle.py
```diff
@@ -31,12 +31,12 @@
     operator_assets_dir = os.path.join("deploy")
```
2uasimojo, line 31: This whole script ought to go away, no?

2uasimojo (commented):
Okay, so the main thing here was the removal of a bunch of yaml files. Is it possible those were intended for testing use? In which case it would be better to preserve them and move them to a different directory.

Makefile
```diff
@@ -3,13 +3,13 @@ include boilerplate/generated-includes.mk
 SHELL := /usr/bin/env bash
 
 # Extend Makefile after here
-CONTAINER_ENGINE?=docker
+CONTAINER_ENGINE=docker
```
lisa, line 6: This is redundant.

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
lisa, line 28: `sort --version-sort` ?

config/templates/csv-template.yaml
bdematte: in the scope of the migration, it is better to be able to still generate the CSV both ways (with common and with hack scripts). once good, we should move the file for good.

2uasimojo: I don't think I agree with that philosophy. You would need to switch between them with a PR anyway, and a revert is just as easy as anything else. And while it *may* be possible in some circumstances/projects to have both mechanisms workable at once, it won't be easy to accomplish; and it will be even harder to support/prove. Then to "complete" the migration we have to do the work anyway.

IMHO it's easy to prove that the migration has worked -- by comparing the bundle files before/after -- so it should be all or nothing.

Makefile
2uasimojo: Worse actually: now it will ignore `CONTAINER_ENGINE` set in the environment.

We should actually just remove this line (though arguably that could be a separate PR) since it's already [set properly and sanely in boilerplate's standard.mk](https://github.com/openshift/boilerplate/blob/f2167d69d378ebb468a49e54f2cae068ab10a383/boilerplate/openshift/golang-osd-operator/standard.mk#L19).

boilerplate/openshift/golang-osd-operator/csv-generate/catalog-build.sh
2uasimojo: TIL. +1.

PR #162, opened 2021-04-09
[WIP] Use shared credentials file instead of static creds

sam-nguyen7:
Implements the AWS SDK's default handling of shared credentials file. Currently, this operator only supports static credentials from the aws_access_key_id and aws_secret_access_key in the aws secret. 

This PR adds in support for another style of authentication by reading 'credentials' key to load shared credential file and use this to authenticate with AWS. The existing  aws_access_key_id and aws_secret_access_key will be included in this shared credentials file.


sam-nguyen7:
Closed in favor of #173

PR #163, opened 2021-04-12
Idle timeout

sedroche:
Jira: https://issues.redhat.com/browse/OSD-6565

Changing the idle timeout to 1800 from 600 on rh-api and default-router loadbalancers.

### DIFF
--- a/pkg/controller/utils/conditions.go
+++ b/pkg/controller/utils/conditions.go
@@ -50,6 +50,7 @@ func SetAPISchemeCondition(
 					Reason:             reason,
 					Message:            message,
 					LastTransitionTime: now,
+					LastProbeTime:      now,
 				},
 			)
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/utils/conditions.go
```diff
@@ -50,6 +50,7 @@ func SetAPISchemeCondition(
 					Reason:             reason,
 					Message:            message,
 					LastTransitionTime: now,
+					LastProbeTime:      now,
```
sedroche, line 53: This was error-ing in the logs. `LastProbeTime` needs a  value set

csheremeta: this sounds like it will fix https://issues.redhat.com/browse/OSD-6789. I'll test this today.

dustman9000:
My only concern with opening up the idle timeout from 10 to 30 minutes is that it has the potential to open up the api endpoint to a slow http request DoS attack. Can we just reduce the backplane timeout to 10m instead?

csheremeta:
I tested this code today to ensure that it fixes both OSD-6565 and OSD-6789 (which it does). I also ran e2e tests against it, and it did not cause any failures.

sedroche:
> My only concern with opening up the idle timeout from 10 to 30 minutes is that it has the potential to open up the api endpoint to a slow http request DoS attack. Can we just reduce the backplane timeout to 10m instead?

@dustman9000 For the rh-api endpoint, that is ip whitelisted. It can only be connected to from the bastions or hives. So its just SRE using it which will mean fewer connections

fahlmant (approved):
/lgtm

PR #168, opened 2021-04-20
Remove unused deploy scripts

jharrington22:
These scripts are no longer used since they have moved to boilerplate.


2uasimojo (commented):
You could also remove [codecov.sh](https://github.com/openshift/cloud-ingress-operator/blob/8f464dbe474e7181ec83894029e891d1dcb5b997/hack/codecov.sh), [validate_yaml.py](https://github.com/openshift/cloud-ingress-operator/blob/8f464dbe474e7181ec83894029e891d1dcb5b997/hack/validate_yaml.py), [generate_template.py](https://github.com/openshift/cloud-ingress-operator/blob/8f464dbe474e7181ec83894029e891d1dcb5b997/hack/generate_template.py)

But

/lgtm

as far as it goes.

PR #164, opened 2021-04-14
Boilerplate update, switch OLM build to boilerplate

jharrington22:
This closes PR #160 as the base of this PR is the boilerplate update including required changes.
 
I've purposely not squashed commits so the PR can be reviewed commit by commit.

I've tested the the catalog registry is returning the correct version when built and below is the diff of bundle files.

```
$ docker run --name catalog-image -d --rm --network host "quay.io/jharrington22/cloud-ingress-operat
or-registry:staging-latest"                                                                                                                                                                                                                                                               
d6052900bc32fc180989eaf70cac96ecc4cc3f5b245425ab9652634edeb3a3df

$ docker run --rm --network=host quay.io/rogbas/grpcurl -plaintext localhost:50051 api.Registry/List
Bundles | jq 'select(.version|test("0532a1e")) | .version'                                                                                                                       
"0.1.346-0532a1e"
```

The following custom resources have been removed from the bundle as they are already deployed in the SelectorSyncSet here
OLM
- CatalogSource
- Subscription
- OperatorGroup
CIO
- APIScheme
- SSHD
- Namespace
COO
- Cloud Credential Requests

Testing

Staging catalog source:
```
docker run --name catalog-image -d --rm --network host "quay.io/jharrington22/cloud-ingress-operator-registry:staging-latest"                      
34bd8e05f048d64b6883a287b82c3e997395606549974cadb9bbfa5223d793b7
$ docker run --rm --network=host quay.io/rogbas/grpcurl -plaintext localhost:50051 api.Registry/ListBundles | jq -r '.version' | sort -V  | tail -n 2
0.1.341-5c00138
0.1.354-c07ed78
```

Production catalog source:
```
$ docker run --name catalog-image -d --rm --network host "quay.io/jharrington22/cloud-ingress-operator-registry:production-latest"                   
bd2e51e02e22f4922251697fe5def6e43d590201520eac2b3978e7b1fe42a198
$ docker run --rm --network=host quay.io/rogbas/grpcurl -plaintext localhost:50051 api.Registry/ListBundles | jq -r '.version' | sort -V  | tail -n 2
0.1.331-10af8da
0.1.354-c07ed78
```

Resulting CSV diff 

[...]

PR #169, opened 2021-04-20
Post-boilerplate cleanup

2uasimojo:
Remove obsolete hack scripts and make targets


jharrington22:
@2uasimojo I didn't remove the `generate_template.py` in #168 because I wanted to see if it was used else where. But it looks like its just a helper script. 

This is what added the CRDs to the Selector Sync Set in the first place.

Before we merge this I want to make sure all clusters get `v0.1.364-9a6ba79` first

2uasimojo:
> @2uasimojo I didn't remove the `generate_template.py` in #168 because I wanted to see if it was used else where. But it looks like its just a helper script.

I'm pretty sure the script relies on the [other resources you deleted in #164](https://github.com/openshift/cloud-ingress-operator/pull/164/files#diff-991950eb890e424406e645aab3e79b280998877aa3c97dfd8488bebdcf8cac19) so it was already dead. If it's still needed, we'll have to resurrect the whole shooting match anyway.

But yeah, there's nothing urgent about this PR -- I actually spun it up in case you needed something to merge yesterday to kick deploys :)

fahlmant (approved):
/hold cancel
/lgtm

fahlmant (approved):
/hold cancel
/lgtm

PR #171, opened 2021-04-23
Change resource requests for operator

cblecker:
Up the memory limit to 4G, remove the CPU limit, and create a CPU request for 200m

Observed usage:

Screen Shot 2021-04-23 at 15 34 20


clcollins (commented):
(This is lgtm, but I guess hold due to code freeze)

wanghaoran1988:
I am surprised that CIO consumes so much mem, could someone explain?

fahlmant (approved):
/lgtm
Approving as a temporary workaround until we can find root cause

PR #172, opened 2021-05-04
OSD-6663: Only UPSERT rh-api/rh-ssh DNS Record if it doesn't exist

bng0y:
This should clean up AWS cloudtrail logs, as it suppresses no-op UPSERTs.

### DIFF
--- a/pkg/cloudclient/aws/private.go
+++ b/pkg/cloudclient/aws/private.go
@@ -643,25 +643,74 @@ func (c *Client) deleteARecord(clusterDomain, DNSName, aliasDNSZoneID, resourceR
 	return err
 }
 
+// recordExists checks if a specific RecordSet already exist in route53
+func (c *Client) recordExists(resourceRecordSet *route53.ResourceRecordSet, publicHostedZoneID string) (bool, error) {
+	if resourceRecordSet == nil {
+		return false, goError.New("resourceRecordSet can't be nil")
+	}
+	if resourceRecordSet.Name == nil {
+		return false, goError.New("resourceRecordSet Name is required")
+	}
+	// Route53 assumes that the domain name that you specify is fully qualified.
+	// Therefore, it doesn't mind if the trailing "." is missing.
+	// However, adding the trailing "." make the future comparaison easier, since route53 listResourceRecordSets contains trailing dots
+	if !strings.HasSuffix(*resourceRecordSet.Name, ".") {
+		*resourceRecordSet.Name += "."
+	}
+	if resourceRecordSet.AliasTarget != nil && !strings.HasSuffix(*resourceRecordSet.AliasTarget.DNSName, ".") {
+		*resourceRecordSet.AliasTarget.DNSName += "."
+	}
+
+	var recordExists bool
+	input := &route53.ListResourceRecordSetsInput{
+		HostedZoneId: aws.String(publicHostedZoneID),
+	}
+
+	// Check if any property of the record was changed.
+	// aws-go-sdk may potentially give all results in multiple pages, so this will go
+	// through every page given in response to the API call
+	err := c.route53Client.ListResourceRecordSetsPages(input, func(p *route53.ListResourceRecordSetsOutput, lastPage bool) bool {
+		for _, record := range p.ResourceRecordSets {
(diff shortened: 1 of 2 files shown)


pkg/cloudclient/aws/private.go
```diff
@@ -648,6 +648,27 @@ func (c *Client) upsertARecord(clusterDomain, DNSName, aliasDNSZoneID, resourceR
...
+				recordExists = true
+				return false
+			}
+		}
+		return true
+	})
+
+	if err != nil || recordExists {
+		return err
+	}
```
lisa, line 671: It would be nice to put this into a method that can be called to "ask" if a given record exists (with a given value?) so that it can be tested in isolation. Speaking of tests, it'd be nice to have tests to exercise the behaviour of this piece of code.

Some of the cases I'm curious about: What is the expected behaviour if the RRTYPE isn't `A`? How does your code change the current behaviour of `upsertARecord` changing the value of the record if it differs from what we want?

lisa (changes requested):

pkg/cloudclient/aws/private.go
bng0y: Thanks for the review, I updated the PR to reflect your comments. 
As for the behaviour when RRTYPE doesn't match (or any other field), the function will return that it didn't find a match and `upsertARecord` will run the ChangeResourceRecordSets UPSERT  as usual.

fahlmant (approved):
/lgtm
/hold

@lisa mind giving this one more look over?

pkg/cloudclient/aws/private.go
```diff
@@ -643,25 +643,72 @@ func (c *Client) deleteARecord(clusterDomain, DNSName, aliasDNSZoneID, resourceR
...
+func (c *Client) recordExists(resourceRecordSet *route53.ResourceRecordSet, publicHostedZoneID string) (bool, error) {
+	if resourceRecordSet == nil {
+		return false, goError.New("resourceRecordSet can't be nil")
+	}
+	if resourceRecordSet.Name == nil {
+		return false, goError.New("resourceRecordSet Name is required")
+	}
+	// Route53 assumes that the domain name that you specify is fully qualified.
+	// Therefore, it doesn't mind if the trailing "." is missing.
+	// However, adding the trailing "." make the future comparaison easier, since route53 listResourceRecordSets contains trailing dots
```
lisa, line 656: Great comment! 👍

pkg/cloudclient/aws/private.go
```diff
@@ -643,25 +643,72 @@ func (c *Client) deleteARecord(clusterDomain, DNSName, aliasDNSZoneID, resourceR
...
+	if resourceRecordSet.AliasTarget != nil && !strings.HasSuffix(*resourceRecordSet.AliasTarget.DNSName, ".") {
+		*resourceRecordSet.AliasTarget.DNSName += "."
+	}
+
+	var recordExists bool
+	input := &route53.ListResourceRecordSetsInput{
+		HostedZoneId: aws.String(publicHostedZoneID),
+	}
+
+	// Check if any property of the record was changed
```
lisa, line 669:
```suggestion
	// Check if any property of the record was changed
	// aws-go-sdk may potentially give all results in multiple pages, so this will go
	//  through every page given in response to the API call
```

Nit: Add a note about pages to match with the following comment in the lambda func to help with reading comprehension.

pkg/cloudclient/aws/private.go
```diff
@@ -643,25 +643,72 @@ func (c *Client) deleteARecord(clusterDomain, DNSName, aliasDNSZoneID, resourceR
...
+			HostedZoneId:         aws.String(aliasDNSZoneID),
+		},
+		Name: aws.String(resourceRecordSetName),
+		Type: aws.String("A"),
+	}
+
+	recordExists, err := c.recordExists(resourceRecordSet, publicHostedZoneID)
+	if err != nil || recordExists {
+		return err
+	}
```
lisa, line 704: Nice concise logic here!

lisa (approved):
Minor nit around an extra comment, but overall looks good. If you like the suggestion, go ahead and apply it, then squash and we can `/lgtm` and `/approve` again. Otherwise, unhold when ready to merge.

pkg/cloudclient/aws/private.go
bng0y: thanks for that

PR #176, opened 2021-05-12
feat(cmd): add MultiNamespacedCacheBuilder to manager

boranx:
Fixes: https://issues.redhat.com/browse/OSD-7118

_Currently, this PR is **inactive/blocked**. Because prior to [v0.8.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.8.0) for controller-runtime, when using cache's client, it's impossible making a _Get_ for a cluster scoped object that doesn't have a namespace._
_https://github.com/kubernetes-sigs/controller-runtime/issues/934_

_This problem seems fixed as of [v0.8.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.8.0)
https://github.com/kubernetes-sigs/controller-runtime/pull/1326_

With MultiNamespacedCacheBuilder, we're making a call to be able to get cluster platform(https://github.com/openshift/cloud-ingress-operator/blob/master/pkg/utils/infrastructure.go#L18-L22), which is a global resource, then we're seeing errors in the reconcile loop of the controllers because of that issue above:
```
"logger":"controller_sshd","msg":"Failed to get cluster's platform","error":"unable to get: /cluster because of unknown namespace for the cache"
```

**Requirement**:

- Upgrade https://github.com/openshift/cloud-ingress-operator/blob/master/go.mod#L26 beyond 0.8.0 (the latest is v0.8.3) which will unblock/ should resolve the issues.


deploy/50_cloud-ingress-operator.Deployment.yaml
```diff
@@ -36,7 +36,7 @@ spec:
           env:
             # "" so that the cache can read objects outside its namespace
             - name: WATCH_NAMESPACE
-              value: ""
+              value: "openshift-sre-sshd,openshift-cloud-ingress-operator,openshift-ingress,openshift-ingress-operator,openshift-kube-apiserver"
```
lisa, line 39: Does this work?? That's awesome, and way better.

boranx: yeah https://github.com/openshift/cloud-ingress-operator/pull/177/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R104 made this available :)

pkg/controller/routerservice/router_service_controller.go
```diff
@@ -48,10 +49,28 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Only filter on services in the openshift-ingress namespace and create/update events
 	p := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return e.Meta.GetNamespace() == RouterServiceNamespace
+			object, err := meta.Accessor(e.Object)
+			if err != nil {
+				return false
```
2uasimojo, line 54: I guess you added this condition because you saw errors happening here? Are those errors worth logging?

2uasimojo (commented):
This looks good from what I can tell. A couple of nonblocking questions/observations inline. I'll leave it to CIO maintainers to stamp.

pkg/cloudclient/aws/private.go
```diff
@@ -129,8 +129,7 @@ func updateAWSLBList(kclient client.Client, oldLBList []awsproviderapi.LoadBalan
 			return err
 		}
 		machineToPatch.Spec.ProviderSpec = *newProviderSpecEncoded
-		machineObj := machineToPatch.DeepCopyObject()
```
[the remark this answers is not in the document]
boranx: the `machineToPatch` itself is being passed to the patch on the next line.
```go
if err := kclient.Patch(context.Background(), &machineToPatch, baseToPatch); err != nil {
```
Why would we need machineObj being created? 

additionally, the version below doesn't compile
```go
machineToPatch.Spec.ProviderSpec = *newProviderSpecEncoded
machineObj := machineToPatch.DeepCopyObject()
if err := kclient.Patch(context.Background(), machineObj, baseToPatch); err != nil {
	log.Error(err, "Failed to update LBs in machine's providerSpec", "machine", machineToPatch.Name)
	return err
}
```
due to `cannot use machineObj (variable of type runtime.Object) as client.Object value in argument to kclient.Patch: missing method GetAnnotations compilerInvalidIfaceAssign`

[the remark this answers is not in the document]
2uasimojo: Ah -- because controller-runtime changed the signature of `Patch`.

> Why would we need machineObj being created?

In theory, if `machineObj` were being passed to this method by reference, or if by-value did a shallow copy, this

```go
machineToPatch.Spec.ProviderSpec = *newProviderSpecEncoded
```

would affect the object in the calling scope, which could have unintended side effects.

pkg/controller/routerservice/router_service_controller.go
boranx: no. looks like `Meta` is no longer an available option for `event.CreateEvent`
`e.Meta undefined (type event.CreateEvent has no field or method Meta)`
`Meta` seems migrated to its own package, the accessor takes the e.Object as param to be able to get its Namespace.
```go
package meta ("k8s.io/apimachinery/pkg/api/meta")
```

However, if this can be refactored to be simplified, happy to refactor here.

fahlmant (approved):
/lgtm
/hold

pkg/cloudclient/aws/private.go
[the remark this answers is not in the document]
boranx: ahh, got your concern now, I've changed it to be using DeepCopy()

pkg/controller/routerservice/router_service_controller.go
2uasimojo: Gotcha.

I think this simplification works:

```go
			return e.Object.GetNamespace() == RouterServiceNamespace
```

boranx: @2uasimojo yeah, agreed, this has been simplified now

pkg/controller/routerservice/router_service_controller.go
```diff
@@ -48,10 +49,28 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Only filter on services in the openshift-ingress namespace and create/update events
 	p := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return e.Meta.GetNamespace() == RouterServiceNamespace
+			object, err := meta.Accessor(e.Object)
```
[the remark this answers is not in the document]
boranx: using the same pattern
```
return e.ObjectNew.GetNamespace() == RouterServiceNamespace
```

fahlmant (approved):
/lgtm
/hold

PR #177, opened 2021-05-12
Update to operator-sdk 0.18.2

cblecker:
This updates the cloud-ingress-operator to use a newer version of the operator-sdk, as well as some other libraries like the operator-custom-metrics.

The main driver for this is to get a slightly newer version of the controller-runtime in support of OSD-7118. This doesn't directly fix that issue, but it's a step in the right direction.


fahlmant (approved):
/lgtm

PR #182, opened 2021-05-25
Update SSS template to use Upsert apply mode

cblecker:
- Update SSS template to use Upsert apply mode
- Remove unused template files

This is step 1 of [OSD-7345](https://issues.redhat.com/browse/OSD-7345)


fahlmant (approved):
/lgtm

PR #184, opened 2021-06-03
Disable cloud-i-o on PrivateLink clusters

dustman9000:
As part of OSD-7433, there is no need for cloud-ingress-operator to be running on PrivateLink clusters. This disables the operator altogether.


jharrington22 (approved):
/lgtm

PR #186, opened 2021-06-05
bug: fix crashloop due to err type mismatch

boranx:
This to be fixing error type conversion mismatch and due to it being in crashloop continuously. 

```
E0605 13:33:11.415451 1086858 runtime.go:78] Observed a panic: &runtime.TypeAssertionError{_interface:(*runtime._type)(0x20f33c0), concrete:(*runtime._type)(0x20754a0), asserted:(*runtime._type)(0x2075440), missingMethod:""} (interface conversion: error is *errors.LoadBalancerNotReadyError, not *errors.DnsUpdateError)
goroutine 1609 [running]:
```


fahlmant (approved):
/lgtm

PR #194, opened 2021-07-14
refactor: add liveness probe to the operator

boranx:
Fix: https://issues.redhat.com/browse/OSD-7407

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
@@ -94,7 +95,8 @@ func main() {
 	}
 
 	options := manager.Options{
-		Namespace: namespace,
+		HealthProbeBindAddress: ":" + osdMetricsPort,
+		Namespace:              namespace,
 	}
 
 	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -182,3 +190,7 @@ func addMetrics(ctx context.Context) {
 		log.Error(err, "Failed to configure OSD metrics")
 	}
 }
+
+func healthCheck(req *http.Request) error {
+	return nil
```
fahlmant, line 195: Is there a case where we want to check the response here? Does this check basically just enable ping?

boranx: for now, it's just enabling ping check which is similar to [healthz.Ping](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.9.6/pkg/healthz/healthz.go#L186-L187) However, we can fill this with validations as a checker function

fahlmant: Cool, makes sense

fahlmant (approved):
/lgtm
/approve
/hold

boranx:
@fahlmant currently it's bound to metric servers port(8181) would we consider having it on a specific port different than that?

fahlmant:
@boranx does the health check pass when testing? As long as the metrics server comes up fast enough then I think it's ok.

boranx:
yup, that's what I see on my local
```shell
curl 127.0.0.1:8181/healthz
ok
```

fahlmant:
I think it's a good starting place then. Feel free to unhold when ready

PR #195, opened 2021-07-20
refactor: error handling, cli generation

boranx:
This PR contains a small number of refactoring items:

- defer `panic()` to outer of the context so we can carry the errors and then remove the panic completely then putting a proper error handling. (only applied for aws)
-  migrate _GetSecrets_ that being used during initialization of the clients to _utils_ and enchant with tests
-  remove sts ruins as it's not being actively used. (? not being sure)
-  add small unit tests to the creation of the client


boranx:
closing this because https://github.com/openshift/cloud-ingress-operator/pull/202 is covering most

PR #196, opened 2021-07-21
[WIP] increasing test coverage

bng0y:


pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -29,3 +44,263 @@ func TestClusterBaseDomain(t *testing.T) {
...
+				cloud: cloud,
+				cobj:  apischemeWithDeleteTimestampWithFinalizer.Obj,
+				robj:  nlbService.Obj,
+				err:   cioerrors.NewDNSUpdateError("bad"),
+			},
+		},
+	}
+
+	for _, test := range tests {
+		testClient, testScheme := testutils.SetUpTestClient(test.ClientObj, test.RuntimeObj, test.ClientErr["on"], test.ClientErr["type"], test.ClientErr["target"])
```
georgettica, line 256: shouldn't the map[string]string be an object of itself and not a generic map?

pkg/controller/apischeme/apischeme_controller_test.go
```diff
@@ -29,3 +44,263 @@ func TestClusterBaseDomain(t *testing.T) {
...
+
+		if err == nil && test.ErrorExpected || err != nil && !test.ErrorExpected {
+			t.Fatalf("Test [%v] return mismatch. Expect error? %t: Return %+v", test.Name, test.ErrorExpected, err)
+		}
+		if err != nil && test.ErrorExpected && test.ErrorReason != fmt.Sprint(k8serr.ReasonForError(err)) {
+			t.Fatalf("Test [%v] FAILED. Excepted Error %v. Got %v", test.Name, test.ErrorReason, k8serr.ReasonForError(err))
+		}
+		if result != test.Resp {
+			t.Fatalf("Test [%v] FAILED. Excepted Response %v. Got %v", test.Name, test.Resp, result)
+		}
```
georgettica, line 278: does that mean that only one error is displayed at a time while running these tests? or would it show the whole suite

this might be related to the fact all of the code is run in one main `TestReconcile`

georgettica: also, this is copied as it's clearer to read this way? I am fine with that but just throwing it out there

georgettica (commented):
I read through the code and had a bit of things to note on them, though adding tests is always good :)

fahlmant (changes requested):
Please provide a description

PR #202, opened 2021-09-20
apply client healthchecking options

boranx:
this adds basic calls to healthchecking for cloudclient. 
the calls can go from straightforward to complex depending on which items is enough to be checked

Tests against AWS cluster:

```shell
curl localhost:8000/healthz
ok
```

Tests against GCP cluster:

```shell
curl localhost:8000/healthz
ok
```

If healthcheck fails, it won't cause the main to be exited instead just returns an http code other than 200 and indirectly causes it to be killed by k8s
```shell
{"level":"info","ts":1633004752.2862713,"logger":"controller-runtime.healthz","msg":"healthz check failed","statuses":[{}]}
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -12,14 +12,15 @@ import (
 
 	operatorconfig "github.com/openshift/cloud-ingress-operator/config"
 	"github.com/openshift/cloud-ingress-operator/pkg/apis"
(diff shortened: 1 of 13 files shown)


fahlmant (approved):
/lgtm
/hold

PR #209, opened 2021-11-03
Limited RBAC for namespaced roles and added required roles to the SSS that can't be deployed by the bundle

fahlmant:
This PR solves two problems:
1. Even after the previous RBAC refactor, some of the roles are too permissive and reference resources that CIO never uses. I've rescoped all the roles related to external namespaces from the operators to have the minimal required permissions
2. Not all roles are being deployed in the OLM bundle and there has been a conflict between the SSS and the OLM bundle with the clusterrole. I've added all the roles that are not deployed by the bundle into the SSS and updated the clusterrole definition to match what we expect.

### DIFF
--- a/deploy/20_cloud-ingress-operator.ClusterRole.yaml
+++ b/deploy/20_cloud-ingress-operator.ClusterRole.yaml
@@ -1,9 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  creationTimestamp: null
   name: cloud-ingress-operator
-  namespace: openshift-cloud-ingress-operator
 rules:
 - apiGroups:
   - config.openshift.io
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -1,372 +1,612 @@
-apiVersion: v1
-kind: Template
-parameters:
-- name: REGISTRY_IMG
-  required: true
-- name: CHANNEL
-  required: true
-- name: IMAGE_TAG
-  required: true
-- name: IMAGE_DIGEST
-  requred: true
-- name: REPO_NAME
-  value: cloud-ingress-operator
-  required: true
-- name: DISPLAY_NAME
-  value: Cloud Ingress Operator
-  required: true
-metadata:
-  name: selectorsyncset-template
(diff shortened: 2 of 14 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -117,58 +117,187 @@ objects:
           skipServiceCheck: true
     - apiVersion: rbac.authorization.k8s.io/v1
       kind: ClusterRole
+      metadata:
+        name: cloud-ingress-operator
+        namespace: openshift-cloud-ingress-operator
```
sam-nguyen7, line 122: can namespace field be removed?

deploy/20_cloud-ingress-operator.ClusterRole.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
```
sam-nguyen7, line 2: do we need to file anymore?

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -117,58 +117,187 @@ objects:
...
+          - get
+          - watch
+      - apiGroups:
+        - config.openshift.io
+        resources:
+        - apiservers
+        verbs:
+        - patch
+        - update
+        - watchkind: ClusterRoleBinding
```
sam-nguyen7, line 141: what is watchkind? do we need it for the roles? or just clusterRole?

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -360,7 +502,7 @@ objects:
       spec:
         targetNamespaces:
         - openshift-${REPO_NAME}
-    - apiVersion: operators.coreos.com/v1alpha1
+    - apiVersion: operators.coreos.com/v1alpha1_
```
sam-nguyen7, line 505: is the _ intentional?

deploy/20_cloud-ingress-operator.ClusterRole.yaml
fahlmant: Yes. This is still the source of truth for what goes in the SSS and the OLM bundle.

PR #215, opened 2021-11-09
OSD-8430/8431 Update Apischeme CR status with every transition + recr…

bng0y:
…eate LB for rh-api if accidentially deleted

- APIscheme CR status will show every transition of the CR instead of the two first currently 
- Loadbalancer used for rh-api will be recreated if accidentally deleted

### DIFF
--- a/deploy/50_cloud-ingress-operator.Deployment.yaml
+++ b/deploy/50_cloud-ingress-operator.Deployment.yaml
@@ -54,4 +54,4 @@ spec:
               scheme: HTTP
               port: 8000
             initialDelaySeconds: 25
-            periodSeconds: 15
+            periodSeconds: 60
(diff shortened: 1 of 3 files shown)


ritmun:
It lgtm, but on a related note, can we add a test for this scenario in osde2e?

deploy/50_cloud-ingress-operator.Deployment.yaml
```diff
@@ -54,4 +54,4 @@ spec:
               scheme: HTTP
               port: 8000
             initialDelaySeconds: 25
-            periodSeconds: 15
+            periodSeconds: 60
```
fahlmant, line 57: What is this used for?

bng0y: This is the time period between each health check. At 15 sec, the pod will crash before the controller can recreate the Load balancer

bng0y:
> It lgtm, but on a related note, can we add a test for this scenario in osde2e?

In my opinion, this should be carded separately

fahlmant (approved):
/lgtm

ritmun:
> > It lgtm, but on a related note, can we add a test for this scenario in osde2e?
> 
> In my opinion, this should be carded separately

In principle, unit tests go with code change, but since we don't have unit tests, what you're saying is ok. 

However, if a code change breaks any tests - e2e or unit tests, it makes sense that the same developer fixes those tests. So we should put e2e tests somewhere in the CI pipeline, if they aren't already.

PR #219, opened 2021-11-19
fix: pagination problem on healthchecking - re-enable healthchecks

boranx:
Fix: https://issues.redhat.com/browse/OSD-9004

cc: @fahlmant @bng0y @tiwillia


fahlmant (approved):
/lgtm

PR #220, opened 2021-11-29
OSD-9056 Inject system trusted-ca-bundle into the running container

mrbarge:
### What type of PR is this?

Feature

### What this PR does / why we need it?

This PR adds introduces a `trusted-ca-bundle` ConfigMap for `cloud-ingress-operator` which shall contain the system's CertificateAuthority bundle  to authenticate TLS connections. [0]

This feature is required in order for CIO to run successfully on OCP clusters that are using a [cluster-wide SSL Forward proxy](https://docs.openshift.com/container-platform/4.9/networking/enable-cluster-wide-proxy.html). Without it, CIO will not trust its communications to `*.amazonaws.com` or `*.googleapis.com`.

The ConfigMap contains the `config.openshift.io/inject-trusted-cabundle` label so that the user-supplied CA bundle is injected into it. More information about this behaviour can be found in the [OCP Documentation](https://docs.openshift.com/container-platform/4.9/networking/configuring-a-custom-pki.html).

The ConfigMap is mounted into CIO in the `/etc/pki/ca-trust/extracted/pem` location, such that no other work is required to have the application use these CA for verifying server cert authenticity.

The approach and naming convention used in this PR aligns with other OCP cluster operators and SRE operators.

### Which Jira/Github issue(s) this PR fixes?

[OSD-9056](https://issues.redhat.com/browse/OSD-9056) 

### Special notes for your reviewer:

- This change has been successfully tested to ensure that OLM can deploy/update the new resource successfully.
- This change has been successfully tested on GCP and AWS clusters in both scenarios of either using or not using a cluster-wide proxy.
- This same change has been implemented in `managed-upgrade-operator` ([PR](https://github.com/openshift/managed-upgrade-operator/pull/292)) and tested successfully with no issues encountered.

### Pre-checks (if applicable):
- [X] Tested latest changes against a cluster
- [X] Ran `make generate` command locally to validate code changes
- [ ] Included documentation changes with PR

### DIFF
--- a/deploy/20_cloud-ingress-operator_trusted-ca-bundle.ConfigMap.yaml
+++ b/deploy/20_cloud-ingress-operator_trusted-ca-bundle.ConfigMap.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-cloud-ingress-operator
+  name: trusted-ca-bundle
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
--- a/deploy/50_cloud-ingress-operator.Deployment.yaml
+++ b/deploy/50_cloud-ingress-operator.Deployment.yaml
@@ -26,6 +26,14 @@ spec:
       - operator: Exists
         key: node-role.kubernetes.io/infra
         effect: NoSchedule
+      volumes:
+      - configMap:
+          defaultMode: 420
+          items:
+            - key: ca-bundle.crt
+              path: tls-ca-bundle.pem
+          name: trusted-ca-bundle
+        name: trusted-ca-bundle
       containers:
         - name: cloud-ingress-operator
           # Replace this with the built image name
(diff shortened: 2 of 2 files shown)


deploy/20_cloud-ingress-operator_trusted-ca-bundle.ConfigMap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
```
fahlmant, line 2: Does this configmap get deployed via the OLM bundle?

mrbarge: That's correct, it does. I've done some tests to verify that OLM is successfully:
- installing it (via a freshly-built and deployed catalog image)
- not clobbering it after cluster-network-operator injects the CAs into it
- cleaning it up if the CSV is removed

fahlmant (approved):
/approve
/hold

Feel free to unhold when ready

PR #222, opened 2021-12-07
Add a tag check when getting the internal API NLB

fahlmant:
https://issues.redhat.com/browse/OSD-9295

This fixes a bug where cloud-ingress-operator can target the wrong NLB when switched from public to private. Currently, when switching schemes, CIO gets a list of all NLBs owned by the cluster (using a tag on the resource). It then loops through that list and finds the first one with scheme == internal. This can lead to the wrong NLB being targeted if there are any Service LoadBalancers with type NLB created within the cluster. To fix this, a check has been added to find the specific internal API NLB that the installer creates. Currently it relies on the naming scheme of OCP, which is <clusterName>-int.

### DIFF
--- a/pkg/cloudclient/aws/private.go
+++ b/pkg/cloudclient/aws/private.go
@@ -880,13 +880,61 @@ func (c *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient 
 				return "", "", err
 			}
 		}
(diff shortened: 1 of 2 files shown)


robotmaxtron (approved):
/lgtm

robotmaxtron (approved):
/lgtm

PR #224, opened 2021-12-09
remove openshift-sre-sshd namespace deployment from CIO 

fahlmant:
Removing deployment of the openshift-sre-sshd namepace from OLM, as its causing issues during build

```
17:49:05   File "./boilerplate/openshift/golang-osd-operator/csv-generate/common-generate-operator-bundle.py", line 130, in <module>
17:49:05     raise UnsupportedRegistryResourceKind(doc['kind'], path)
17:49:05 __main__.UnsupportedRegistryResourceKind: The resource at deploy/20_cloud-ingress-operator._openshift-sre-ssh.Namespace.yaml of kind Namespace is not supported
```

### DIFF
--- a/deploy/20_cloud-ingress-operator._openshift-sre-ssh.Namespace.yaml
+++ b/deploy/20_cloud-ingress-operator._openshift-sre-ssh.Namespace.yaml
@@ -1,3 +1,0 @@
-kind: Namespace
-metadata:
-  name: openshift-sre-sshd


jharrington22 (approved):
/lgtm

jharrington22:
@fahlmant apply mode is sync this will remove the namespace from all clusters. Set it to Upsert and we'll promote it through and set it back to sync later.

jharrington22 (approved):
/lgtm

PR #229, opened 2021-12-10
Add patch and update to machine object during public private swtich

fahlmant:
When switching from private to public, the machine objects gets patched to match the list of NLBs they're registered with. Currently this permission is missing, so while the LBs and DNS gets updated correctly, the Machine CRs are not representative of the actual config.


sam-nguyen7 (approved):
/lgtm

ritmun (approved):

PR #230, opened 2021-12-13
Remove last excessive permissions from clusterrole

fahlmant:
Remove last excessive permissions from the operator.


robotmaxtron (approved):
/lgtm

PR #232, opened 2022-01-12
role fix to allow service deletion in openshift-kube-apiserver namespace

bng0y:
Allows  cio's to delete rh-api service in openshift-kube-apiserver for recovery when the associated LB was manually deleted

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -332,6 +332,7 @@ objects:
         - watch
         - create
         - update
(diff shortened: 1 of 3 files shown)


fahlmant (changes requested):
This role definition also needs to be updated in the selectorsyncset here https://github.com/openshift/cloud-ingress-operator/blob/master/hack/olm-registry/olm-artifacts-template.yaml#L318

fahlmant (approved):
/approve

fahlmant (approved):
/lgtm

PR #238, opened 2022-02-26
Osd 10155

ritmun:
CIO can detect missing  rh-api LB in GCP and recreate it.

To test, create a GCP CCS cluster and follow this SOP:  https://docs.google.com/document/d/1QsZrxifiLwzbi_rC1Lz1Av9HcgXAzwXnvoAJtoPTEZo/edit   

[I'll squash commits when review updates are done.]

### DIFF
--- a/pkg/cloudclient/gcp/gcp.go
+++ b/pkg/cloudclient/gcp/gcp.go
@@ -28,8 +28,8 @@ var (
 	log = logf.Log.WithName("gcp_cloudclient")
 )
 
-// GCPClient represents a GCP cloud Client
-type GCPClient struct {
+// Client represents a GCP cloud Client
+type Client struct {
 	projectID      string
 	region         string
 	clusterName    string
@@ -40,37 +40,37 @@ type GCPClient struct {
 }
 
 // EnsureAdminAPIDNS implements cloudclient.CloudClient
-func (c *GCPClient) EnsureAdminAPIDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
+func (c *Client) EnsureAdminAPIDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
 	return c.ensureAdminAPIDNS(ctx, kclient, instance, svc)
 }
 
 // DeleteAdminAPIDNS implements cloudclient.CloudClient
-func (c *GCPClient) DeleteAdminAPIDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
+func (c *Client) DeleteAdminAPIDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
 	return c.deleteAdminAPIDNS(ctx, kclient, instance, svc)
 }
 
 // EnsureSSHDNS implements cloudclient.CloudClient
-func (c *GCPClient) EnsureSSHDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.SSHD, svc *corev1.Service) error {
+func (c *Client) EnsureSSHDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.SSHD, svc *corev1.Service) error {
 	return c.ensureSSHDNS(ctx, kclient, instance, svc)
 }
 
 // DeleteSSHDNS implements cloudclient.CloudClient
-func (c *GCPClient) DeleteSSHDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.SSHD, svc *corev1.Service) error {
+func (c *Client) DeleteSSHDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.SSHD, svc *corev1.Service) error {
 	return c.deleteSSHDNS(ctx, kclient, instance, svc)
 }
 
 // SetDefaultAPIPrivate implements cloudclient.CloudClient
-func (c *GCPClient) SetDefaultAPIPrivate(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.PublishingStrategy) error {
+func (c *Client) SetDefaultAPIPrivate(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.PublishingStrategy) error {
 	return c.setDefaultAPIPrivate(ctx, kclient, instance)
 }
 
 // SetDefaultAPIPublic implements cloudclient.CloudClient
-func (c *GCPClient) SetDefaultAPIPublic(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.PublishingStrategy) error {
+func (c *Client) SetDefaultAPIPublic(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.PublishingStrategy) error {
 	return c.setDefaultAPIPublic(ctx, kclient, instance)
 }
 
 // Healthcheck performs basic calls to make sure client is healthy
-func (c *GCPClient) Healthcheck(ctx context.Context, kclient client.Client) error {
+func (c *Client) Healthcheck(ctx context.Context, kclient client.Client) error {
 	out, err := c.computeService.RegionBackendServices.List(c.projectID, c.region).Do()
 	if err != nil {
 		return err // possible client deformation
@@ -91,7 +91,7 @@ func performHealthCheck(l *computev1.BackendServiceList, clusterName string) err
 	return fmt.Errorf("internal lb not found: exiting to refresh")
 }
 
-func newClient(ctx context.Context, serviceAccountJSON []byte) (*GCPClient, error) {
+func newClient(ctx context.Context, serviceAccountJSON []byte) (*Client, error) {
 	credentials, err := google.CredentialsFromJSON(
 		ctx, serviceAccountJSON,
 		dnsv1.NdevClouddnsReadwriteScope,
@@ -110,7 +110,7 @@ func newClient(ctx context.Context, serviceAccountJSON []byte) (*GCPClient, erro
 		return nil, err
 	}
 
-	return &GCPClient{
+	return &Client{
 		projectID:      credentials.ProjectID,
 		dnsService:     dnsService,
 		computeService: computeService,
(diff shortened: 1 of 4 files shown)


fahlmant:
I see this is in draft. Is this ready for review.

ritmun:
> I see this is in draft. Is this ready for review?
Yep, it shouldn't have been a draft. I'm going to squash all commits after review, before actual merge so we can review this.

pkg/cloudclient/gcp/private.go
```diff
@@ -115,10 +115,28 @@ func (c *Client) setDefaultAPIPublic(ctx context.Context, kclient client.Client,
...
 }
 
-func (c *Client) ensureDNSForService(kclient client.Client, svc *corev1.Service, dnsName string) error {
+func (c *GCPClient) ensureDNSForService(kclient client.Client, svc *corev1.Service, dnsName string) error {
 	// google.golang.org/api/dns/v1.Service is a struct, not an interface, which
 	// will make this all but impossible to write unit tests for
 
+	// Forwarding rule is necessary for rh-api lb setup
+	// Check forwarding rule exists first
+	if svc.Spec.Type == "LoadBalancer" {
```
fahlmant, line 124: Is there ever a case where this this service won't be LoadBalancer? If so, should there be an error?

pkg/cloudclient/gcp/private.go
```diff
@@ -115,10 +115,28 @@ func (c *Client) setDefaultAPIPublic(ctx context.Context, kclient client.Client,
...
+		ingressList := svc.Status.LoadBalancer.Ingress
+		if len(ingressList) == 0 {
+			// the LB doesn't exist
+			return cioerrors.NewLoadBalancerNotReadyError()
+		}
+		rhapiLbIP := ingressList[0].IP
```
fahlmant, line 130: Will the target IP always be in the first item?

pkg/cloudclient/gcp/private.go
```diff
@@ -115,10 +115,28 @@ func (c *Client) setDefaultAPIPublic(ctx context.Context, kclient client.Client,
...
+		// ensure forwarding rule exists in GCP for service
+		_, err := findGCPForwardingRuleForExtIP(c, rhapiLbIP)
+		if err != nil {
+			return cioerrors.ForwardingRuleNotFound(err.Error())
+		}
+	} else {
+		log.Info("service type is not LoadBalancer; will not ensure forwarding rule in GCP.")
```
fahlmant, line 137: Should there be a return here?

pkg/cloudclient/gcp/gcp.go
```diff
@@ -28,8 +28,8 @@ var (
 	log = logf.Log.WithName("gcp_cloudclient")
 )
 
-// Client represents a GCP Client
-type Client struct {
+// GCPClient represents a GCP cloud Client
+type GCPClient struct {
```
fahlmant, line 32: This change is slightly out of scope. Can this be spun out into a separate refactoring card?

pkg/cloudclient/gcp/private.go
```diff
@@ -187,7 +205,26 @@ func (c *Client) ensureDNSForService(kclient client.Client, svc *corev1.Service,
...
+	for _, lb := range response.Items {
+		if lb.IPAddress == rhapiLbIP {
+			fr = lb
+		}
+	}
+	if fr == nil {
+		return nil, fmt.Errorf("Forwarding rule not found in GCP for given service IP %s", rhapiLbIP)
+	}
+	return fr, nil
+}
```
fahlmant, line 225:
```suggestion
func (c *GCPClient) findGCPForwardingRuleForExtIP(rhapiLbIP string) (compute.ForwardingRule, error) {
	listCall := c.computeService.ForwardingRules.List(c.projectID, c.region)
	response, err := listCall.Do()
	if err != nil {
		return nil, err
	}

	for _, lb := range response.Items {
		if lb.IPAddress == rhapiLbIP {
			return lb, nil
		}
	}
	
	return compute.ForwardingRule{}, nil
}
```

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -257,12 +260,29 @@ func (r *ReconcileAPIScheme) Reconcile(ctx context.Context, request reconcile.Re
...
 		r.SetAPISchemeStatusMetric(instance)
 		return reconcile.Result{}, err
+	case *cioerrors.ForwardingRuleNotFoundError:
+		// This error handles the missing/deleted forwarding rule/LB in cloud provider
+		r.SetAPISchemeStatus(instance, "Couldn't reconcile", "Forwarding rule was deleted on cloud provider", cloudingressv1alpha1.ConditionError)
+
+		// To recover from this case we will need to delete the lb service.
+		// It will be recreated  at the next reconcile.
+		reqLogger.Info(fmt.Sprintf("Forwarding rule was deleted on cloud provider, deleting service %s/service/%s to force recreation", found.GetNamespace(), found.GetName()))
+		err1 := r.client.Delete(context.TODO(), found)
```
fahlmant, line 276: minor nit, can err1 be more explicitly named?

fahlmant (changes requested):

pkg/cloudclient/gcp/private.go
```diff
@@ -115,10 +115,28 @@ func (c *Client) setDefaultAPIPublic(ctx context.Context, kclient client.Client,
...
+	// Check forwarding rule exists first
+	if svc.Spec.Type == "LoadBalancer" {
+		ingressList := svc.Status.LoadBalancer.Ingress
+		if len(ingressList) == 0 {
+			// the LB doesn't exist
+			return cioerrors.NewLoadBalancerNotReadyError()
+		}
+		rhapiLbIP := ingressList[0].IP
+		// ensure forwarding rule exists in GCP for service
+		_, err := findGCPForwardingRuleForExtIP(c, rhapiLbIP)
```
fahlmant, line 132: There should also be a path to handle if the returned value is an empty struct.

pkg/cloudclient/gcp/private.go
ritmun: I'm checking for the type to be on the safer side, and not throwing an error, because I don't want to interfere with any existing functionality for those services.

pkg/cloudclient/gcp/private.go
ritmun: I relied on Benson's existing logic to get hostname from the first item. This item has both hostname and IP. @bng0y  might know the answer to whether it's always the first item.

pkg/cloudclient/gcp/private.go
fahlmant: The apischeme controller creates this Service, so we can rely on the code used to create the service there. If there's a type check, then each type should be explicitly handled.

pkg/cloudclient/gcp/gcp.go
ritmun: Done https://github.com/openshift/cloud-ingress-operator/pull/240

pkg/cloudclient/gcp/private.go
ritmun: I'm not sure about other types and it seems too out of scope, so I'll remove this check and I can create a separate card for handling types here.

ritmun: done

pkg/cloudclient/gcp/private.go
ritmun: I'm not able to commit it as suggestion here, so I'll push the update.

ritmun: done

pkg/cloudclient/gcp/private.go
ritmun: findGCPForwardingRuleForExtIP will return nil instead of an empty struct since its return value is a pointer. But that still needed handling. Added missing check. Thanks for spotting that.

pkg/cloudclient/gcp/private.go
```diff
@@ -187,6 +202,22 @@ func (c *Client) ensureDNSForService(kclient client.Client, svc *corev1.Service,
 	return nil
 }
 
+// Returns GCP forwarding rule for given IP
+func findGCPForwardingRuleForExtIP(c *Client, rhapiLbIP string) (*compute.ForwardingRule, error) {
```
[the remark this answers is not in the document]
ritmun: Updated in new commit. Couldn't commit suggestion as is, because it accompanied change in the function call.

pkg/cloudclient/gcp/private.go
```diff
@@ -129,8 +129,8 @@ func (c *Client) ensureDNSForService(kclient client.Client, svc *corev1.Service,
 	}
 	rhapiLbIP := ingressList[0].IP
 	// ensure forwarding rule exists in GCP for service
-	fr, err := findGCPForwardingRuleForExtIP(c, rhapiLbIP)
-	if err != nil || fr == nil {
+	_, err := c.findGCPForwardingRuleForExtIP(rhapiLbIP)
```
fahlmant: The nil case still needs to be handled since nil is returned in line 218

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -96,6 +98,7 @@ type LoadBalancer struct {
 // Rough Steps:
 // 1. Create Service
 // 2. Add DNS CNAME from rh-api to the ELB created by AWS provider
+// 3. Add Forwarding rule in GCP for the lb service
 // 3. Ready for work (Ready)
```
fahlmant, line 102:
```suggestion
// 4. Ready for work (Ready)
```

pkg/controller/apischeme/apischeme_controller.go
```diff
@@ -257,12 +260,29 @@ func (r *ReconcileAPIScheme) Reconcile(ctx context.Context, request reconcile.Re
...
 		r.SetAPISchemeStatus(instance, "Couldn't reconcile", "Couldn't ensure the admin API endpoint: "+err.Error(), cloudingressv1alpha1.ConditionError)
 		r.SetAPISchemeStatusMetric(instance)
 		return reconcile.Result{}, err
+	case *cioerrors.ForwardingRuleNotFoundError:
+		// This error handles the missing/deleted forwarding rule/LB in cloud provider
+		r.SetAPISchemeStatus(instance, "Couldn't reconcile", "Forwarding rule was deleted on cloud provider", cloudingressv1alpha1.ConditionError)
+
+		// To recover from this case we will need to delete the lb service.
+		// It will be recreated  at the next reconcile.
+		reqLogger.Info(fmt.Sprintf("Forwarding rule was deleted on cloud provider, deleting service %s/service/%s to force recreation", found.GetNamespace(), found.GetName()))
```
fahlmant, line 275: This error only affects GCP, so I think its safe to say GCP rather than cloud proivder

fahlmant (approved):
/lgtm

PR #239, opened 2022-03-04
feat: add query function for clusterversion

boranx:
Pre-step&feat: https://issues.redhat.com/browse/OSD-10213

This PR contains a helper function for querying up-to-date cluster version.

Usage:
 
```go
// Get cluster version
cv, err := utils.GetClusterVersionObject(r.client)
if err != nil {
}
reqLogger.Info(cv.Status.Desired.Version) // print/evaluate
```


fahlmant (commented):
I think we need to go one step further and set a "global" version var once this resource has been returned

boranx:
I'd agree to go one step further and add such combinations to decide if the version is higher than 4.10 (as it was the main need for most cards) 
However, I've actually concerns about setting this variable as global: wouldn't this cause ambiguity on the clusters if they upgrade the cluster from 4.9 to a 4.10+? We'd have to restart the operator to refresh the global variable.
I didn't get why we don't go with call-when-needed way

fahlmant:
If a cluster gets upgraded, all nodes should be rebooted which means all pods should get rebooted. I want to avoid doing this call every reconcile loop; if we can prevent that then we can make it on-dmeand.

boranx:
agreed :+1: if the upgrade won't be an issue, let's do it :) I'll move it to WIP till the work is done on this PR

fahlmant (commented):
/lgtm

PR #240, opened 2022-03-07
Osd 10229 - rename  GCP Client struct 

ritmun:
Currently this struct name "Client" conflicts with kubeclient and causes confusion in some places. Rename this to clarify the struct distinction.

### DIFF
--- a/pkg/cloudclient/gcp/gcp.go
+++ b/pkg/cloudclient/gcp/gcp.go
@@ -28,8 +28,8 @@ var (
 	log = logf.Log.WithName("gcp_cloudclient")
 )
 
-// Client represents a GCP Client
-type Client struct {
+// GCPClient represents a GCP Cloud Client
+type GCPClient struct {
 	projectID      string
 	region         string
 	clusterName    string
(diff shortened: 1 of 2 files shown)


rafael-azevedo:
@ritmun not sure if this is necessary. Since the package name is called gcp when this client is created it would be called `gcp.Client`. If we rename it to gcpclient it would be redundent `gcp.gcpclient`

ritmun:
> @ritmun not sure if this is necessary. Since the package name is called gcp when this client is created it would be called `gcp.Client`. If we rename it to gcpclient it would be redundant `gcp.gcpclient`

It is necessary for disambiguation. There are places in the code where two different Client structs are used in the same place creating much ambiguity. (Attaching screenshot.) This is not desirable for readability. 
image

I think this change makes code more readable.

boranx:
I'd agree with using `gcp.gcpclient` would sound a bit repeating. Would it makes sense to change client.Client to something like `k8s.Client`?

ritmun:
> 

I like this. I'll push an update.

ritmun:
> I'd agree with using `gcp.gcpclient` would sound a bit repeating. Would it makes sense to change client.Client to something like `k8s.Client`?

 

I'd also like to change gcp.Client to gcp.CloudClient since CloudClient is the accurate name of the struct we're implementing.

ritmun:
I've applied suggested change. I'm in process of squashing all commits and making a new PR.

PR #249, opened 2022-03-17
Osd 10229 new

ritmun:
Alias k8s client and rename client vars to differentiate cloudclient instances

### DIFF
--- a/pkg/cloudclient/gcp/gcp.go
+++ b/pkg/cloudclient/gcp/gcp.go
@@ -16,7 +16,7 @@ import (
 	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	k8s "sigs.k8s.io/controller-runtime/pkg/client"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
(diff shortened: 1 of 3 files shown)


pkg/cloudclient/gcp/gcp.go
```diff
@@ -40,42 +40,42 @@ type Client struct {
 }
 
 // EnsureAdminAPIDNS implements cloudclient.CloudClient
-func (c *Client) EnsureAdminAPIDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
-	return c.ensureAdminAPIDNS(ctx, kclient, instance, svc)
+func (gcpclient *Client) EnsureAdminAPIDNS(ctx context.Context, kclient k8s.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
```
boranx, line 43: Hey Ritu,

Using a single letter for representing the parent struct for bound functions is common (but not a convention in general I guess, I couldn't find official docs but saw similar usages everywhere).
to me, I like the pattern because it looks clean and prevents me from typing a bound object's name again and again. 
Therefore, I'm not with changing the name with a longer version but it's just an opinion. I'd ping @fahlmant  and ask his opinion on it

1- https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/cookie.go;l=178
2- https://github.com/operator-framework/operator-sdk/blob/master/internal/olm/client/client.go#L112

rafael-azevedo: +1 I this is a convention we use across all code bases current and a standard in go. I think we are safe to keep things as is currently.

ritmun: I don't know why that's a convention. It was very confusing to write this code with multiple struct names starting with "c" and in addition 2 structnames actually named "client". What's the logic here?

ritmun:
continued here https://github.com/openshift/cloud-ingress-operator/pull/251/files

PR #251, opened 2022-03-28
Osd 10229 7

ritmun:
Jira: https://issues.redhat.com/browse/OSD-10229 
Refactor: disambiguate multiple usages of "Client" for readability

### DIFF
--- a/pkg/cloudclient/gcp/gcp.go
+++ b/pkg/cloudclient/gcp/gcp.go
@@ -16,7 +16,7 @@ import (
 	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	k8s "sigs.k8s.io/controller-runtime/pkg/client"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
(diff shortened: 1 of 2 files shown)


pkg/cloudclient/gcp/gcp.go
```diff
@@ -40,39 +40,39 @@ type Client struct {
 }
 
 // EnsureAdminAPIDNS implements cloudclient.CloudClient
-func (c *Client) EnsureAdminAPIDNS(ctx context.Context, kclient client.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
-	return c.ensureAdminAPIDNS(ctx, kclient, instance, svc)
+func (gc *Client) EnsureAdminAPIDNS(ctx context.Context, kclient k8s.Client, instance *cloudingressv1alpha1.APIScheme, svc *corev1.Service) error {
```
boranx, line 43: looks ok, we can merge this if the consensus
Still, I'd like to share my thoughts: I think still my vote is that there can be different package clients which can have the same name(just client).
[1]a package is a collection of source files in the same directory that are compiled together. Functions, types, variables, and constants defined in one source file are visible to all other source files within the same package, but they won't be available from other packages so we can say packages are isolated in their context,  in other saying _aws.client_ != _gcp.client_

I think naming functions for obvious definitions is important, yes, but we can also consider that packages (source files being separate folders) also creating a meaning separation.

1- https://go.dev/doc/code#Code%20organization

ritmun: Hey Boran, thanks for your comment. This isn't only to differentiate from AWS client, but also to differentiate from kubeclient. It's a perspective from a new person working on this code: the overuse of word "client" is dizzying in this file. This will help it.

boranx: so why don't we focus on differentiating kubeclient just and let the rest of it to a discussion with team? I'm convinced that we should simplify and prevent repeating of using `client` word but modifying struct identifier is not something I'd prefer in my honest opinion

```go
func NewClient(k8sCli k8s.Client) (*Client, error) {
```

ritmun: Is changing c to gc such a great disadvantage to warrant further discussion? I've described its merit, I don't have any further arguments to make to others.

boranx: oki, if we feel confident with this, then I'm, too 
feel free to close this :)

PR #253, opened 2022-04-05
rbac rules added to launch sshd pod 

reetika-vyas:
Jira: This is for OSD-10573.
Added RBAC  rules to launch a sshd pod inside a private cluster to access their addon UI


fahlmant:
We should not be using sshd pods in external clusters

cc @cblecker

fahlmant:
/hold
For further review

cblecker:
We won't be expanding SSHD pods, and will actually be pulling them back soon.

/close

PR #255, opened 2022-04-09
OSD-10308 added unit test for ensure forwarding rule

ritmun:
Maintenance sprint PR. Added unit test to cover recently added GCP code. 
Jira:  https://issues.redhat.com/browse/OSD-10308

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,12 +13,14 @@ require (
 	github.com/openshift/cluster-api-provider-gcp v0.0.0
 	github.com/openshift/machine-api-operator v0.2.1-0.20200226185612-9b0170a1ba07
 	github.com/openshift/operator-custom-metrics v0.4.2
+	github.com/openshift/origin v0.0.0-20160503220234-8f127d736703
 	github.com/operator-framework/operator-sdk v0.18.2
 	github.com/prometheus/client_golang v1.7.1
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/oauth2 v0.0.0-20200902213428-5d25da1a8d43
 	google.golang.org/api v0.35.0
 	gopkg.in/yaml.v2 v2.3.0
+	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.20.2
 	k8s.io/apimachinery v0.20.2
 	k8s.io/client-go v12.0.0+incompatible
--- a/pkg/cloudclient/gcp/gcp_test.go
+++ b/pkg/cloudclient/gcp/gcp_test.go
@@ -1,16 +1,22 @@
 package gcp
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/openshift/cloud-ingress-operator/config"
 	"github.com/openshift/cloud-ingress-operator/pkg/testutils"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	"gotest.tools/assert"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-func TestNewClient(t *testing.T) {
+var cli *Client
+
+func setupSuite(t *testing.T) {
 	dummySA := `{
 		"type": "service_account",
 		"private_key_id": "abc",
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -13,12 +13,14 @@ require (
 	github.com/openshift/cluster-api-provider-gcp v0.0.0
 	github.com/openshift/machine-api-operator v0.2.1-0.20200226185612-9b0170a1ba07
 	github.com/openshift/operator-custom-metrics v0.4.2
+	github.com/openshift/origin v0.0.0-20160503220234-8f127d736703
```
fahlmant, line 16: What's pulling this in?

pkg/cloudclient/gcp/gcp_test.go
```diff
@@ -1,16 +1,22 @@
...
+	"gotest.tools/assert"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-func TestNewClient(t *testing.T) {
+var cli *Client
+
+func setupSuite(t *testing.T) {
```
fahlmant, line 19: This is currently not being called anywhere.

ritmun: Yes, the setup and tearsdown functions are implicitly called, which lint doesn't seem to consider. We'll need to make a lint exception here. Not sure how that's done, will have to find out.

ritmun: Ref: https://coreos.slack.com/archives/CMD1L32M9/p1649693572235999?thread_ts=1649685655.318599&cid=CMD1L32M9

fahlmant: Why was the name of the function changed?

ritmun: Before, creating the client was the test. Now creating the client is a part of set up so other tests in that suite can use it globally. I couldn't kept the old test, but it would be redundant if the setup was to do the same thing.

fahlmant: I'm not following how setupSuite is implicitly called then.

ritmun: I thought it was, but it wasn't. I went down the rabbit hole a bit to make it call setupSuite implicitly, but that would need me to change the test to use the testify Suite struct, which is more than we want to do here. 

Reverted this to as before, moved the test to private_go.

go.mod
ritmun: Removed this by go mod tidy

pkg/cloudclient/gcp/private.go
```diff
@@ -128,7 +130,7 @@ func (gc *Client) ensureDNSForService(kclient k8s.Client, svc *corev1.Service, d
 	}
 	rhapiLbIP := ingressList[0].IP
 	// ensure forwarding rule exists in GCP for service
-	err := gc.ensureGCPForwardingRuleForExtIP(rhapiLbIP)
+	err := gc.ensureGCPForwardingRuleForExtIP(rhapiLbIP, getForwardingRuleList)
```
fahlmant, line 133: I'm not clear why we're passing a function as a parameter for this. Why can't `ensureGCPForwardingRuleForExtIp` call `getForwardingRuleList` directly?

ritmun: This follows higher order function mocking. It's one of the ways of mocking a function so that we cover precisely the function we're testing.  The higher order function is private and is kept extremely simple to isolate the external dependency. https://medium.com/@tahasevim231/mocking-with-high-order-functions-in-go-b76d5a3103d2

fahlmant: Looking at this again, I'm not convinced getForwadingRuleList needs to be its own function at this point. Its only ever called once and its not a method of the gcp Client but instead takes it as an argument so I don't see much value in it being a function that's only available in this package as it current stands

ritmun: It's for mocking the unit test correctly. This is a common method of unit testing. It's increasing coverage, so it's good for the code overall.

fahlmant: This changes the ergonomics of the code and makes it fairy complicated going forward but with no functional change, except to make mocking tests simpler. We have other functions with a similar pattern that would also all need to be refactored to this pattern if we want to test them as well.

IMO, we should instead abstract out the GCP API client from the gc Client to an interface with the list of required methods. That way we can pass in a mocked or "real" implementation to the gc Client struct.

PR #258, opened 2022-05-11
Create local definition of IngressController

fahlmant:
In order to use some of the newer features of IngressControllers without
directing using the openshift/api 4.10 release branch, this commit
copies the IngressController definitions locally. Only the parts of
the spec needed are imported, while the rest are dropped for simplicity.


jharrington22:
@fahlmant so far looks reasonable to me so far

rafael-azevedo:
looks good @fahlmant . So this means we will need to update this anytime the definition changes going forward ?

fahlmant:
This is a temporary measure until we can update the operator-sdk/controller-runtime/api versions. Note that the IngressController definition here is already a subset of the 4.10 openshift/api definition. So we only have to add fields that we want (as long as they're marked optional) and can ignore the others.

boranx:
public&private testing is successful against a cluster having version: _4.10.12_

boranx (commented):
/lgtm

PR #260, opened 2022-05-13
[OSD-7396] Validating webhooks

bng0y:
Adding a webhook to prevent deletion of protected apischeme and sshd CRs. Protecting the resources is done through annotation: `managed.openshift.io/protected-delete: "true"`

To test the change:
- Annotate the apischeme `oc annotate apischeme/rh-api managed.openshift.io/protected-delete="true"`
- Attempt to delete it `oc delete apischeme/rh-api`
- Deletion attempt should fail


fahlmant:
@bng0y are you still planning on working on this?

PR #263, opened 2022-06-21
Remove everything SSHD related

fahlmant:
Related to https://issues.redhat.com//browse/OSD-11724 and https://github.com/openshift/managed-cluster-config/pull/1163

Hive is the only place we're still using SSHD. Since we're removing it from hive, this PR removes everything related from CIO deployments.


fahlmant:
/hold

This needs to be synchronized with https://github.com/openshift/managed-cluster-config/pull/1163 to avoid the operator from erroring on existing clusters

cblecker:
@fahlmant I think there are still some ssh references in `pkg/cloudclient/aws/private_test.go` and `pkg/testutils/testutils.go`. Would you mind having a look at thise?

PR #264, opened 2022-07-12
OSD-11442 - Address gosec lint warnings

anispate:
Reassigned the value of the machine in the for loop to avoid "G601: Implicit memory aliasing in for loop" error,  add a false positive to the AWSSecretName and GCPSecretName to avoid "G101: Potential hardcoded credentials" error and added package "golang.org/x/sys/unix".


fahlmant (approved):
/approve
/lgtm

PR #265, opened 2022-08-10
Set ELBConnectionTimeout on IngressController for v4.11+

blrm:
As of OCP v4.11, cluster-ingress-operator allows management of AWS ELB idle connection timeouts by setting them in the spec of the IngressController: https://github.com/openshift/cluster-ingress-operator/pull/451 . Since cloud-ingress-operator currently sets that idle timeout by annotating the LoadBalancer service object directly, it was causing a fight between the two operators were cloud-ingress-operator was applying the annotation to the service and cluster-ingress-operator was removing it, causing frequent AWS calls as the setting was modified back and forth.

This PR updates the publishingstrategy controller to set the ELB idle timeout directly in the IngressController for v4.11+ clusters, and to patch any IngressControllers owned by cloud-ingress-operator that lack that field in their spec. 

I tested this against 4.10 and 4.11 clusters. On 4.10, the current behavior still worked correctly and the cloud-ingress-operator makes no attempt to set the connectionIdleTimeout on the IngressController. On 4.11, it correctly patched the default IngressController to set the idle connection timeout there, and once set it no longer competes with the cluster-ingress-operator to set the annotation on the LB service object.

### DIFF
--- a/pkg/controller/publishingstrategy/publishingstrategy_controller.go
+++ b/pkg/controller/publishingstrategy/publishingstrategy_controller.go
@@ -34,6 +34,7 @@ const (
 	infraNodeLabelKey          = "node-role.kubernetes.io/infra"
 	CloudIngressFinalizer      = "cloudingress.managed.openshift.io/finalizer-cloud-ingress-controller"
 	ClusterIngressFinalizer    = "ingresscontroller.operator.openshift.io/finalizer-ingresscontroller"
+	ELBIdleTimeoutDuration     = 1800
 )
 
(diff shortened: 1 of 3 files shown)


fahlmant (approved):
/lgtm
/approve

I'm going to add
/hold
as I'd like at least one person from Aurora to also sign off :)

lnguyen1401:
Thank you for fixing this !

blrm:
Thanks @lnguyen1401 and @fahlmant! Now that Linh has lgtm'ed as well, I'm going to drop the hold.

/unhold

PR #267, opened 2022-09-05
[OSD-11790] Migrate to operator-sdk v1.x

bmeng:
Generally migration to operator-sdk v1.x and run the boilerplate makes.

pkg/controller/ -> controllers/
pkg/apis -> api/
cmd/manager/main.go -> main.go

Additional changes:
- pkg/controller/utils -> pkg/controllerutils
- fixed an empty machineList issue in aws/provider_test.go


controllers/apischeme/apischeme_controller.go
```diff
@@ -384,3 +363,10 @@ func sliceEquals(left, right []string) bool {
 	}
 	return true
 }
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *APISchemeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Owns(&cloudingressv1alpha1.APIScheme{}).
```
mrbarge, line 370: Are we sure this is the right one, and not for example `For`? In the original controller it uses `EnqueueRequestForObject` rather than `EnqueueRequestForOwner` and CIO does not set an owner reference on apischemes.

controllers/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -775,11 +756,18 @@ func (r *ReconcilePublishingStrategy) ensureIngressController(reqLogger logr.Log
...
 	}
 	// If the CR was created, requeue PublishingStrategy
 	return reconcile.Result{Requeue: true}, nil
 
 }
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PublishingStrategyReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Owns(&v1alpha1.PublishingStrategy{}).
```
mrbarge, line 771: Similarly to my earlier comment, we should check that this should not be `For` instead.

main.go
```diff
@@ -0,0 +1,168 @@
...
+	utilruntime.Must(awsproviderapi.SchemeBuilder.AddToScheme(scheme))
+	//+kubebuilder:scaffold:scheme
+}
+
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
```
mrbarge, line 78: The previous CIO defaults this to `8000`. I'm unsure if there's a specific reason but for before/after consistency's sake it may be worth considering changing this.

main.go
```diff
@@ -0,0 +1,168 @@
...
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
```
mrbarge, line 89: The old CIO performs leader-for-life election using `Become()`, we may need to consider if we are maintaining that behaviour.

main.go
```diff
@@ -0,0 +1,168 @@
...
+	// setup publishingstrategycontroller with mgr
+	if err = (&publishingstrategycontroller.PublishingStrategyReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "PublishingStrategy")
+		os.Exit(1)
+	}
+
```
mrbarge, line 121: Adding the router service controller to the manager seems to have been missed.

controllers/apischeme/apischeme_controller.go
bmeng: Yeah, I think it should be `For` instead of `Owns`. 

Fixed, thanks!

controllers/publishingstrategy/publishingstrategy_controller.go
bmeng: Yeah, I think it should be `For` instead of `Owns`. 

Fixed, thanks!

main.go
bmeng: I think so... the routerservice controller added to the manager now

main.go
bmeng: The leader election was added.

main.go
bmeng: This is important... the `8000` is written in the deployment. And if we use the different port in the code. I will result that the health check failure and the operator exit with healthCheck failure... https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L423

mrbarge:
This PR has now passed functional testing performed in OSD-12974, so it can be considered safe for final review/acceptance.

controllers/apischeme/apischeme_controller.go
```diff
@@ -94,19 +70,22 @@ type LoadBalancer struct {
 	BaseDomain   string // What is the base domain (DNS zone) for the EndpointName record?
 }
 
-// Reconcile will ensure that the rh-api management api endpoint is created and ready.
```
cblecker, line 97: Are these old comments still relevant?

controllers/publishingstrategy/publishingstrategy_controller.go
```diff
@@ -48,59 +61,26 @@ var IngressControllerEndPoint patchField = "IngressControllerEndpoint"
...
-	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
+// PublishingStrategyReconciler reconciles a PublishingStrategy object
+type PublishingStrategyReconciler struct {
+	Client client.Client
+	Scheme *runtime.Scheme
 }
 
-// Reconcile reads that state of the cluster for a PublishingStrategy object and makes changes based on the state read
```
cblecker, line 90: Same here.. is this still relevant?

controllers/routerservice/routerservice_controller.go
```diff
@@ -27,68 +40,27 @@ const (
...
-
-// ReconcileRouterService reconciles a Service object
-type ReconcileRouterService struct {
-	// This client, initialized using mgr.Client() above, is a split client
-	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
-}
-
-// Reconcile reads that state of the cluster for a RouterService object and makes changes based on the state read
```
cblecker, line 81: And one more

pkg/cloudclient/aws/private_test.go
```diff
@@ -134,6 +137,8 @@ func TestRemoveAWSELB(t *testing.T) {
 			if err != nil {
 				t.Fatalf("Couldn't reload the test machine (named %s): %v", machineInfo.Name, err)
 			}
+			newMachines = append(newMachines, machine)
```
cblecker, line 140: Curious why we need to do this when we didn't before?

bmeng: https://github.com/kubernetes-sigs/controller-runtime/blame/15b6689c55745514d6bade4c55efb3333c0da4d9/pkg/client/fake/client.go#L358 There is a change in FakeKubeClient(after migrate) which will clear the value in the old machine pointer. 
So it is not needed in the existing code, but will get blank machineList with the new controller-runtime's fakeClient after `Get()`

controllers/apischeme/apischeme_controller.go
bmeng: Thanks, all the comments for the controllers are added back.

bmeng:
@boranx @fahlmant @cblecker @sam-nguyen7 
Any more comments on this? Or can we get this merged?

cblecker:
/lgtm
/approve
/hold

@bmeng Please feel free to remove the hold when you're ready for this to merge (and validate in stage)

fahlmant (approved):
/approve as well :)

PR #270, opened 2022-09-27
osd-10308 adding unit test coverage for GCP forwarding rule 

ritmun:
Changes:

1. Code coverage for GCP code for ensureGCPForwardingRuleForExtIP function
2. Added GCPComputeClient to mock GCP client's GetForwardingRuleList function
3. Added auto generated mock for GCP in pkg/cloudclient/mocks/mock_gcp.go

To test, run `make test` or run individual unit test `TestEnsureGCPForwardingRuleForExtIP` in `pkg/cloudclient/gcp/private_test.go`

Optional: To end to end test forwarding rule functionality, refer to description of the [enhancement PR](https://github.com/openshift/cloud-ingress-operator/pull/238)


pkg/cloudclient/gcp/gcp.go
```diff
@@ -28,15 +28,22 @@ var (
 	log = logf.Log.WithName("gcp_cloudclient")
 )
 
+type GCPComputeClient interface {
+	GetForwardingRuleList() (*computev1.ForwardingRuleList, error)
+}
+
 // Client represents a GCP cloud Client
+// To recreate mock for this, run:
+// mockgen -source=pkg/cloudclient/gcp/gcp.go -destination=pkg/cloudclient/mocks/mock_gcp.go
```
mjlshen, line 37: This is a totally different repo, so not necessarily a good idea, but should we add a Makefile target to generate these mocks like: https://github.com/openshift/osd-network-verifier/blob/main/Makefile#L24-L28 ?

ritmun: I like that! I'll try it out. Also, just realized the path to destination isn't right..

fahlmant:
Copying over comment from previous version of this PR https://github.com/openshift/cloud-ingress-operator/pull/255#discussion_r850715788

```
This changes the ergonomics of the code and makes it fairy complicated going forward but with no functional change, except to make mocking tests simpler. We have other functions with a similar pattern that would also all need to be refactored to this pattern if we want to test them as well.

IMO, we should instead abstract out the GCP API client from the gc Client to an interface with the list of required methods. That way we can pass in a mocked or "real" implementation to the gc Client struct.
```

ritmun:
> Copying over comment from previous version of this PR [#255 (comment)](https://github.com/openshift/cloud-ingress-operator/pull/255#discussion_r850715788)
> 
> This changes the ergonomics of the code and makes it fairy complicated going forward but with no functional change, except to make mocking tests simpler. We have other functions with a similar pattern that would also all need to be refactored to this pattern if we want to test them as well.
> 
> IMO, we should instead abstract out the GCP API client from the gc Client to an interface with the list of required methods. That way we can pass in a mocked or "real" implementation to the gc Client struct.

@fahlmant I'm surprised to see the same comment again, because I scrapped the methods from #255 completely and implemented what was asked for. For reference, your suggestion was 
 "Using interfaces is a common pattern in go and are easier to use" and that's what is implemented here https://coreos.slack.com/archives/CMD1L32M9/p1650390158081969?thread_ts=1650387229.978099&cid=CMD1L32M9

ritmun:
For 

> > Copying over comment from previous version of this PR [#255 (comment)](https://github.com/openshift/cloud-ingress-operator/pull/255#discussion_r850715788)
> > This changes the ergonomics of the code and makes it fairy complicated going forward but with no functional change, except to make mocking tests simpler. We have other functions with a similar pattern that would also all need to be refactored to this pattern if we want to test them as well.
> > IMO, we should instead abstract out the GCP API client from the gc Client to an interface with the list of required methods. That way we can pass in a mocked or "real" implementation to the gc Client struct.
> 
> @fahlmant I'm surprised to see the same comment again, because I scrapped the methods from #255 completely and implemented what was asked for. For reference, your suggestion was "Using interfaces is a common pattern in go and are easier to use" and that's what is implemented here https://coreos.slack.com/archives/CMD1L32M9/p1650390158081969?thread_ts=1650387229.978099&cid=CMD1L32M9

This change follows the pattern both Rafael and you are alluding to in your comments, per the example Rafael suggested  (his link was moved but I found the relevant path in the project) https://github.com/openshift/gcp-project-operator/blob/master/controllers/projectclaim/projectclaim_controller.go

PR #272, opened 2022-09-29
Set cluster version environment variable

mrbarge:
## What

Restores a missing logic block that sets the `CLUSTER_VERSION` environment variable in `main.go`.

## Why

This code was missing in the Operator SDK migration #267 and as a result is causing CIO to perform heavy API load on cloud providers by constantly annotation the `router-default` service.

### DIFF
--- a/main.go
+++ b/main.go
@@ -123,6 +123,13 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


jharrington22:
Nice @mrbarge thanks!

PR #275, opened 2022-12-13
Set external traffic policy for rh-api endpoint

cblecker:
Sets the `.spec.externalTrafficPolicy` for the rh-api load balancer to local. This ensures that rh-api traffic cannot route through a bad worker node, but instead will only pass healthchecks on the nodes that actually are hosting a healthy api server.


cblecker:
/hold
This is ready for review, but I want to control when it merges.

boranx (commented):
/lgtm

putting as a ref: https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
Nice tweak! Feel free to unhold anytime when ready to merge

cblecker:
I have tested the rollout of this on stage, and it was successful.

PR #277, opened 2022-12-13
Remove deprecated ioutil usage

cblecker:
```
SA1019: "io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.
```

This removes use of this deprecated stdlib module.


fahlmant (commented):
/lgtm
/approve

PR #285, opened 2023-05-15
[WIP]disable controller when running on cluster >= 4.13

bmeng:
wrt [OSD-16007](https://issues.redhat.com//browse/OSD-16007)


bmeng:
Tested on both 4.12 and 4.13 clusters. Works well.

fahlmant:
Are we also deprecating api availability? This CR handles both Ingress and API availability

bmeng:
That's good point, API endpoint was not considered, I will need to rework

hectorakemp:
@bmeng probably time to close this?

PR #292, opened 2023-07-13
Use correct hive SSS patch syntax

hectorakemp:
Fixes an issue where the `SelectorSyncSet` patch that was used to let the Cloud Ingress Operator know that a cluster has the `legacy ingress` label was malformed. 

Correct syntax recommended [here](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1689178206664249?thread_ts=1689148277.662989&cid=CE3ETN3J8)), and implemented in this PR.

Completely tested and confirmed to work with a stage cluster and `ClusterDeployment` [here](

This has now been tested using a test selectorsyncset [here](https://docs.google.com/document/d/1ViR9VdvEU9XWPanwluDV8qHZtPHloqY8KU9AQ63SbOw/edit) and confirmed to work.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -449,9 +449,14 @@ objects:
         operator: In
         values:
           - "false"
-    resourceApplyMode: Sync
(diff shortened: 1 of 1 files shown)


boranx (commented):
/lgtm

@hectorakemp thanks for the fix!

PR #301, opened 2023-09-14
Deprecate feature flag in favour of empty applicationIngress on 4.13+

hectorakemp:
The new 'Managed Ingress' feature is triggered globally by the OCM api by setting the label `legacy-ingress-support: false` on the ClusterDeployment. It then uses this label as a boolean to alter the 'PublishingStrategy' Syncset to make `applicationIngress` an empty slice (to reflect that all ingresses have been returned to the cluster ingress operator).

## Issue

Previously, we were activating the feature by setting the same label on the publishing strategy custom resource, which activated the 'new ingress' code path inside the Publishing Strategy reconciler.

The way we set this is by using a SyncSet to pass down the feature label state from the cluster deployment. 

Unfortunately, due to a race condition between these two SyncSets, the `applicationIngress` slice can be set to empty *before* the feature label is set. This can result in the default IngressController being deleted completely from the cluster, losing all previous settings. 

This is documented in two bugs found by QE: (https://issues.redhat.com/browse/OSD-18383, https://issues.redhat.com/browse/OSD-18384). What happens here is that clusters that were set to NLB, or private, or any other state have their ICs deleted by the Publishing Strategy reconciler *before* the syncset can pass the label down. 

## Solution

We only ever send an empty slice of `applicationIngresses` when the feature label for managed ingress v2 is set (`legacy-ingress-support: false`). Assume then that setting application ingress to an empty slice by the OCM API is the way to 'signal' this new feature. 

A byproduct of this is no longer needing the labeller `SyncSet` or any mention of the `ClusterDeployment` feature flag in the codebase. 

Tested+working in stage.


fahlmant (approved):
/lgtm
/approve

PR #303, opened 2023-10-08
[OSD-19012] add annotation to the rh-api service to set the tag for SG

bmeng:
We need the redhat managed Security Group to have the tag "red-hat-managed=true" like the other Loadbalancer services.


mjlshen (approved):
/lgtm

boranx (commented):
/lgtm

PR #309, opened 2023-10-24
Update dependencies in preparation for adding CPMS support

AlexVulaj:
In order to support coming changes for CPMS, we needed to update some of our dependencies. This caused a whole mess of dependency issues that required @JoelSpeed 's help to untangle, which I believe is worthy of its own PR.

Most of this work has been superseded by https://github.com/openshift/cloud-ingress-operator/pull/310, but I think there's still some good prep work here for [OSD-14568](https://issues.redhat.com/browse/OSD-14568).

### DIFF
--- a/pkg/utils/machines.go
+++ b/pkg/utils/machines.go
@@ -4,19 +4,26 @@ package utils
 
 import (
 	"context"
+	"fmt"
+	machinev1 "github.com/openshift/api/machine/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	machineapi "github.com/openshift/api/machine/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const masterMachineLabel string = "machine.openshift.io/cluster-api-machine-role"
+const (
+	masterMachineLabel  = "machine.openshift.io/cluster-api-machine-role"
+	machineApiNamespace = "openshift-machine-api"
+	cpmsName            = "cluster"
+)
 
 // GetMasterMachines returns a MachineList object whose .Items can be iterated
 // over to perform actions on/with information from each master machine object.
 func GetMasterMachines(kclient client.Client) (*machineapi.MachineList, error) {
 	machineList := &machineapi.MachineList{}
 	listOptions := []client.ListOption{
-		client.InNamespace("openshift-machine-api"),
+		client.InNamespace(machineApiNamespace),
 		client.MatchingLabels{masterMachineLabel: "master"},
 	}
 	err := kclient.List(context.TODO(), machineList, listOptions...)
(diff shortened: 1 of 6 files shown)


AlexVulaj:
Not really sure what's going on with these builds. I can run the following locally just fine:
```
$ golangci-lint run -c boilerplate/openshift/golang-osd-operator/golangci.yml ./...
$ 
```
But in the lint ci check, I'm seeing:
```
level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: buildir: failed to load package cache: could not load export data: no export data for \"sigs.k8s.io/controller-runtime/pkg/cache\"\n\n"
```

Does anyone have any idea what's going on here?

JoelSpeed:
Does your [base image](https://github.com/openshift/cloud-ingress-operator/blob/master/.ci-operator.yaml), which I believe is the image these jobs are running with, have an up to date go version?

AlexVulaj:
@JoelSpeed the error ended up being Go versions indeed, with the lint error here being a bit misleading.

Updating our CI images for our operators gets a bit more complicated since we build those images from https://github.com/openshift/boilerplate , and going to the required Go version here (1.20) would be a bit more effort. @mjlshen opened up a PR here that should solve those build issues, and I'll be rebasing off of that once it's merged in, or closing this PR altogether. https://github.com/openshift/cloud-ingress-operator/pull/310

pkg/utils/machines.go
```diff
@@ -25,3 +32,19 @@ func GetMasterMachines(kclient client.Client) (*machineapi.MachineList, error) {
 	}
 	return machineList, nil
 }
+
+func GetControlPlaneMachineSet(kclient client.Client) (*machinev1.ControlPlaneMachineSet, error) {
+	cpms := &machinev1.ControlPlaneMachineSet{}
```
mjlshen, line 37: We would need to register `machinev1` to the scheme in `main.go` as well

pkg/utils/machines.go
```diff
@@ -25,3 +32,19 @@ func GetMasterMachines(kclient client.Client) (*machineapi.MachineList, error) {
...
+	key := client.ObjectKey{
+		Namespace: machineApiNamespace,
+		Name:      cpmsName,
+	}
+	err := kclient.Get(context.TODO(), key, cpms)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil // Nothing to do
+		}
```
mjlshen, line 46: This behavior kinda depends on the use-case for this function, I would prefer to see why this function is needed first before we implement the function.

pkg/utils/machines.go
AlexVulaj: Good call - I think that got overwritten in the rebase.

pkg/utils/machines.go
AlexVulaj: It's just a helper to prep for https://issues.redhat.com/browse/OSD-14568. Would you like me to delete this function until that ticket is fully implemented? (at which point I'd rather just close this PR entirely)

sam-nguyen7: +1 to remove this function until we can understand (and review) it with the full logic flow

rafael-azevedo: +1 to remove this function as well. It should be merged in relevant PR should it be necessary

AlexVulaj:
Superseded by #310

pkg/utils/machines.go
AlexVulaj: This work was initially going to be for the whole ticket, but after running into so many dependency issues I decided to split out this PR separately. Michael's PR #310 supersedes that effort, so there's no real reason to keep this PR around. Going to close it.

PR #314, opened 2023-11-15
[OSD-14568] Handle CPMS enabled / inactive clusters

bergmannf:
Add support to handle CPMS clusters.

In case of an inactive CPMS - remove the CPMS, modify the machines, recreate the CPMS.
In case of an active CPMS - modify the CPMS.

### DIFF
--- a/pkg/cloudclient/aws/private.go
+++ b/pkg/cloudclient/aws/private.go
@@ -855,13 +855,6 @@ func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient
 		return "", "", err
 	}
 	removalClosure := getLoadBalancerRemovalFunc(ctx, kclient, masterList, cpms)
-	if cpms.Spec.State == machinev1.ControlPlaneMachineSetStateInactive {
-		err := baseutils.RemoveCPMSAndAwaitMachineRemoval(ctx, kclient, cpms)
-		if err != nil {
-			log.Error(err, "Failed removing CPMS")
-			return "", "", err
-		}
-	}
 	var intDNSName, intHostedZoneID, lbName string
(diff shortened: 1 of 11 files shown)


bergmannf:
Holding this until I have tested it with an active CPMS

pkg/cloudclient/aws/private.go
```diff
@@ -1099,3 +1110,110 @@ func encodeAWSMachineProviderSpec(awsProviderSpec *machinev1beta1.AWSMachineProv
...
+			// E1117 14:42:49.349992       1 controller.go:255] fbergman-14568-tn7t8-master-0: failed to delete machine: fbergman-14568-tn7t8-master-0: reconciler failed to Delete machine: failed to updated update load balancers: LoadBalancerNotFound: Load balancers '[fbergman-14568-tn7t8-ext]' not found
+			log.Info("Updating controlplane machines")
+			err = removeLoadBalancerMachineSet(ctx, kclient, lbName, masterList)
+			if err != nil {
+				log.Error(err, "failed to remove load balancer from machines")
+				return err
+			}
+			go func() {
+				maxRetries := 5
+				for {
```
AlexVulaj, line 1173:
```suggestion
				for maxRetries := 0; maxRetries < 5; maxRetries++ {
```

bergmannf: I kept the infinite loop as that makes it easier to handle the 'it did not succeed -> print a log' case.
If I use a normal loop I would have to name it and break it to differentiate the cases:
* Broke the loop because success
* Broke the loop because all attempts expired

boranx (commented):
/lgtm
/approve

Nice implementation 👍

PR #316, opened 2023-12-14
Add new CPMS related permissions to the olm deplyoment

fahlmant:
Fixes SA missing permissions from changes in https://github.com/openshift/cloud-ingress-operator/pull/314

Currently, the permissions put in the resources folder must be manually copied to the OLM manifests. This aligns the two so that the permissions will part of the role when OLM deploys the operator.


sam-nguyen7 (approved):
/lgtm
/approve

PR #324, opened 2024-04-24
Fix issue with CIO deleting all NLBs when targeting API NLBs

luis-falcon:
attempt Fix issue with CIO deleting all NLBs when targeting API NLBs:

I noticed that the new behavior left 1 of the 2 internet facing nlbs (only the -ext is deleted) but then again when setting API to public it would not create the ext elb since the code only looked for any "internet-facing" nlb scheme on aws. I fixed that by putting an extra condition to look for "internet-facing" with name that contains "-ext" on it (which was marked as TODO on the code also). 

A quick note on testing, while on stage with an image with operator code when you change the API to private since for this testing all syncsets are paused and other stuff you have to change the publishing strategy to internal and then on uhc console -> network tag set the API to private, after like 15 mins backplane will pick it up and you will be able to log into the cluster again but for a bit you wont be able to log in. After like 15 mins that dns propagates you can log in you can set the publishing strategy to "externa"l and on uhc console the api to public again. On aws you should notice the network load balancer named $CLUSTERNAME-somehash-ext being deleted when internal, recreated when setting to internal.

### DIFF
--- a/pkg/cloudclient/aws/private.go
+++ b/pkg/cloudclient/aws/private.go
@@ -185,9 +185,9 @@ func (ac *Client) setDefaultAPIPublic(ctx context.Context, kclient k8s.Client, i
 	if err != nil {
 		return err
 	}
-	// TODO: Check for the expected name?
+
 	for _, networkLoadBalancer := range nlbs {
-		if networkLoadBalancer.scheme == "internet-facing" {
+		if networkLoadBalancer.scheme == "internet-facing" && strings.HasSuffix(networkLoadBalancer.loadBalancerName, "-ext") {
 			// nothing to do
 			return nil
 		}
@@ -838,6 +838,11 @@ func (ac *Client) ensureDNSRecordsRemoved(clusterDomain, DNSName, aliasDNSZoneID
 
 // removeLoadBalancerFromMasterNodes
 func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient k8s.Client) (string, string, error) {
+	clusterName, err := baseutils.GetClusterName(kclient)
+	if err != nil {
+		return "", "", err
+	}
+
 	nlbs, err := ac.listOwnedNLBs(kclient)
 	if err != nil {
 		return "", "", err
@@ -868,17 +873,29 @@ func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient
 	var intDNSName, intHostedZoneID, lbName string
 	for _, networkLoadBalancer := range nlbs {
 		if networkLoadBalancer.scheme == "internet-facing" {
-			lbName = networkLoadBalancer.loadBalancerName
-			err := ac.deleteExternalLoadBalancer(networkLoadBalancer.loadBalancerArn)
-			if err != nil {
-				return "", "", err
-			}
-			err = removalClosure(lbName)
-			if err != nil {
-				return "", "", err
-			}
-		}
 
+			lbName = networkLoadBalancer.loadBalancerName
+
+			canDelete, err := ac.canDeleteNlb(networkLoadBalancer, clusterName)
+			if err != nil {
+				log.Error(err, "Problem attempting to remove", "NLB", networkLoadBalancer.loadBalancerName)
+				return "", "", err
+			}
+
+			if canDelete {
+
+				log.Info("Removing from cluster", "NLB", networkLoadBalancer.loadBalancerName)
+				err = ac.deleteExternalLoadBalancer(networkLoadBalancer.loadBalancerArn)
+				if err != nil {
+					return "", "", err
+				}
+				err = removalClosure(lbName)
+				if err != nil {
+					return "", "", err
+				}
+			}
+
+		}
 	}
 
 	internalAPINLB, err := ac.getInteralAPINLB(kclient)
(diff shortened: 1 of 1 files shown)


pkg/cloudclient/aws/private.go
```diff
@@ -868,8 +868,20 @@ func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient
...
+			var nlbsToDeleteArn string
+			nlbTagsMap, err := ac.getAllTagsFromLoadBalancer(networkLoadBalancer.loadBalancerArn)
+			if err != nil {
+				fmt.Printf("There was a problem getting back network load balancer tags for NLB %s: %s\n", networkLoadBalancer.loadBalancerArn, err)
+				return "", "", err
+			}
+
 			lbName = networkLoadBalancer.loadBalancerName
-			err := ac.deleteExternalLoadBalancer(networkLoadBalancer.loadBalancerArn)
+			if _, ok := nlbTagsMap["kubernetes.io/service-name"]; !ok {
```
fahlmant, line 880: What if there are multiple clusters in the same AWS account and region? Is there anyway to make sure the LB matches the specific cluster?

luis-falcon: there is a label that gives us the cluster name. I think that should be a good hook to identify the proper cluster.

pkg/cloudclient/aws/private.go
```diff
@@ -868,8 +868,20 @@ func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient
...
+				nlbsToDeleteArn = networkLoadBalancer.loadBalancerArn
+			}
+
+			err = ac.deleteExternalLoadBalancer(nlbsToDeleteArn)
```
bergmannf, line 884: This can be called with an empty String if you never hit your `if` - not sure what will happen in that case - did you verify that?

Might be better to handle the empty string case directly before relying on AWS to do what we hope it does.

pkg/cloudclient/aws/private.go
```diff
@@ -868,17 +873,29 @@ func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient
 	var intDNSName, intHostedZoneID, lbName string
 	for _, networkLoadBalancer := range nlbs {
 		if networkLoadBalancer.scheme == "internet-facing" {
+
 			lbName = networkLoadBalancer.loadBalancerName
-			err := ac.deleteExternalLoadBalancer(networkLoadBalancer.loadBalancerArn)
+
+			canDelete, err := ac.canDeleteNlb(networkLoadBalancer, clusterName)
 			if err != nil {
+				fmt.Printf("Error: Problem while trying to Delete NLB %v: %v \n", networkLoadBalancer.loadBalancerName, err)
```
bergmannf, line 881: I think it might be better to stick with using `log` calls instead of `fmt` - same as in the rest of the file.

pkg/cloudclient/aws/private.go
```diff
@@ -868,17 +873,29 @@ func (ac *Client) removeLoadBalancerFromMasterNodes(ctx context.Context, kclient
...
 				return "", "", err
 			}
-			err = removalClosure(lbName)
-			if err != nil {
-				return "", "", err
+
+			if canDelete {
+
+				fmt.Printf("The following elb will be Removed from cluster \"%v\": { %v }\n", clusterName, networkLoadBalancer.loadBalancerName)
```
bergmannf, line 887: Same here - might be better to use a `log` instead of a `fmt`.

pkg/cloudclient/aws/private.go
luis-falcon: Done

pkg/cloudclient/aws/private.go
luis-falcon: Done

pkg/cloudclient/aws/private.go
```diff
@@ -1219,3 +1236,55 @@ func getLoadBalancerRemovalFunc(ctx context.Context, kclient k8s.Client, masterL
...
+		}
+	}
+
+	return tags, nil
+}
+
+// canDeleteNlb is a function that checks if and nlb can be deleted by checking certain conditions (if certain tags are present)
+func (ac *Client) canDeleteNlb(networkLoadBalancer loadBalancerV2, clusterName string) (bool, error) {
+
+	var belongstoCluster, NotServiceAttached, isExternal bool
```
bergmannf, line 1268:
```suggestion
	var belongstoCluster, notServiceAttached, isExternal bool
```

pkg/cloudclient/aws/private.go
```diff
@@ -1219,3 +1236,55 @@ func getLoadBalancerRemovalFunc(ctx context.Context, kclient k8s.Client, masterL
...
+		return false, err
+	}
+
+	//check if the elb belongs to the right cluster (this is in case there are more than one cluster in the same region)
+	if _, ok := nlbTagsMap["kubernetes.io/cluster/"+clusterName]; ok {
+		belongstoCluster = true
+	}
+	//check the elb is not attached to a service
+	if _, ok := nlbTagsMap["kubernetes.io/service-name"]; !ok {
+		NotServiceAttached = true
```
bergmannf, line 1282:
```suggestion
		notServiceAttached = true
```

pkg/cloudclient/aws/private.go
```diff
@@ -1219,3 +1236,55 @@ func getLoadBalancerRemovalFunc(ctx context.Context, kclient k8s.Client, masterL
...
+	}
+	//check the elb has the external suffix
+	if strings.HasSuffix(nlbTagsMap["Name"], "-ext") {
+		isExternal = true
+	}
+
+	return (belongstoCluster && NotServiceAttached && isExternal), nil
```
bergmannf, line 1289:
```suggestion
	return (belongstoCluster && notServiceAttached && isExternal), nil
```

fahlmant (commented):
One minor nit; the check if the NLB is owned by the cluster and is the Public API LB is separate in `setDefaultAPIPublic` and `removeLoadBalancersFromMasterNodes`. If one of those checks get changed, the other will need to be changed. Combining these checks to use one function would be safer

PR #331, opened 2024-06-17
Sdcicd-885 Update bp

ritmun:
add boilerplate convention for osde2e

will be followed up with test migrations

[SDCICD-885](https://issues.redhat.com//browse/SDCICD-885)


fahlmant (approved):
/lgtm
/approve

PR #332, opened 2024-06-17
Migrate rh-api LB reconciliation e2e tests from osde2e

ritmun:
- Adds rh-api LB e2e tests from osde2e repo (existing test for ref: https://github.com/openshift/osde2e/blob/main/pkg/e2e/operators/cloudingress/rhapi_lb.go) 
- Retains test specs and logic for AWS and GCP
- removes dependency on osde2e repo
- improves error and log messages

[SDCICD-885](https://issues.redhat.com//browse/SDCICD-885)


fahlmant (approved):
/lgtm
/approve

PR #336, opened 2024-07-05
skip rh-api test for sts clusters

ritmun:
1) During test migration, the block that skips rh-api lb e2e tests on sts clusters was lost. Adding it back. Test will be skipped on sts clusters which are used in all rosa nightly e2e test jobs. 

2) osde2e-common library now provides cluster configs such as STS, cloud region, cloud provider; so they don't need to be read from environment variables. That change is incorporated for all 3 variables in test. 

Caused by [SDCICD-885](https://issues.redhat.com//browse/SDCICD-885)


fahlmant (approved):
/lgtm
/approve

PR #340, opened 2024-08-14
OSD-24978: Not deploying CIO on GCP WIF clusters

Nikokolas3270:
Lets make sure CIO (Cloud Ingress Operator) is not deployed on OSD clusters running on GCP with WIF (Workload Identity Federation) enabled.
WIF is the GCP equivalent of AWS STS; both are implementing OAuth 2.0. Our policy in that managed clusters should only deploy base operators (i.e. operators that come with OCP) when OAuth 2.0 is in use; indeed we want to keep to a minimal the actions eventually performed by the cluster on the customer cloud provider account when this authentication scheme is in use.
This excludes CIO which is:
- Not a "base" operator, that is to say not an operator bundled with OCP
- Accessing the customer cloud provider

This PR is related thos this one: https://github.com/openshift/managed-cluster-config/pull/2189


typeid (approved):
/lgtm

fahlmant (approved):
/lgtm
/approve

fahlmant (approved):
/lgtm
/approve

PR #344, opened 2024-09-18
 increase timeout for loadbalancer e2e test to 10 min

ritmun:
rh-api load balancer may take longer than 5 minutes. increasing it to 10 minutes in AWS e2e test, this also matches the test for GCP.


fahlmant (approved):
/lgtm
/approve

PR #345, opened 2024-10-25
OSD-26173: don't deploy CIO on PSC clusters

typeid:
https://issues.redhat.com/browse/OSD-26173

CIO should not be deployed on PSC clusters, as those should not have a `rh-api` LB, similarly to privatelink.


boranx (commented):
/lgtm

PR #352, opened 2025-03-13
Ensure cidr blocks have been completely updated before continuing

btoll:
https://issues.redhat.com/browse/OSD-28341

The unit test is mutating the number of cidr blocks, so we need to wait each time we update the ingress to ensure that the resource has been updated in the cluster before moving on.

It's better to do this in a `wait` function where we can explicitly check for any expectation(s) rather than in a timeout or a defer.


osde2e/cloud_ingress_operator_tests.go
```diff
@@ -202,9 +208,6 @@ var _ = ginkgo.Describe("cloud-ingress-operator", ginkgo.Ordered, func() {
 
 			ginkgo.By("Initializing AWS ELB service")
 			lb := elb.New(awsSession)
-			input := &elb.DeleteLoadBalancerInput{
-				LoadBalancerName: aws.String(oldLBName),
-			}
```
btoll, line 207: Instead of creating a one-off variable, I moved this into the function call below.

osde2e/cloud_ingress_operator_tests.go
```diff
@@ -102,32 +103,22 @@ var _ = ginkgo.Describe("cloud-ingress-operator", ginkgo.Ordered, func() {
...
-		// Wait 30 secs for apiserver to reconcile
-		time.Sleep(30 * time.Second)
+		err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 5*time.Minute, false, func(ctx context.Context) (bool, error) {
+			if slices.Equal(updatedApiScheme.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks, updatedCidrBlock) {
+				log.Println("Updated cidrblock in rh-api service.")
+				return true, nil
+			}
+			return false, nil
+		})
+		Expect(err).NotTo(HaveOccurred(), "Updated cidrblock not able to be set in rh-api service.")
```
btoll, line 121: Removing the arbitrary `sleep` provides more stability for this test and also improves the run time.

osde2e/cloud_ingress_operator_tests.go
```diff
@@ -102,32 +103,22 @@ var _ = ginkgo.Describe("cloud-ingress-operator", ginkgo.Ordered, func() {
 		cidrBlock := apiScheme.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks
 		updatedApiScheme := apiScheme.DeepCopy()
 
-		//reset cidrblock after test is done
-		defer func() {
-			updatedApiScheme.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks = apiScheme.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks
-			err = k8s.Update(ctx, updatedApiScheme)
-			Expect(err).NotTo(HaveOccurred(), "Could not revert APIScheme CR instance")
-		}()
```
btoll, line 110: The logic in this block is moved to the end of the function.

osde2e/cloud_ingress_operator_tests.go
```diff
@@ -138,6 +129,21 @@ var _ = ginkgo.Describe("cloud-ingress-operator", ginkgo.Ordered, func() {
...
+
+		ginkgo.By("Restoring the CIDR block to its original state.")
+		err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 5*time.Minute, false, func(ctx context.Context) (bool, error) {
+			if slices.Equal(updatedApiScheme.Spec.ManagementAPIServerIngress.AllowedCIDRBlocks, cidrBlock) {
+				log.Println("Original cidrblock from apischeme successfully restored in rh-api service.")
+				return true, nil
+			}
+			return false, nil
+		})
+		Expect(err).NotTo(HaveOccurred(), "Original cidrblock not restored in APIScheme.")
```
btoll, line 146: This is the fix for the unit test.  Blocking until the state has been restored improves the stability of the test suite.

Amarthya-v (approved):

osde2e/cloud_ingress_operator_tests.go
btoll: It's not a 5 min timeout, that's the deadline timeout, so if it doesn't succeed within 5 mins it will fail.  It usually succeeds < 30s.

ktrainor7 (approved):

joshbranham (commented):
/lgtm

PR #364, opened 2025-06-08
SREP-96 Updates to selectorSyncSet to prevent stuck subscriptions

luis-falcon:
Updates to selectorSyncSet to prevent stuck subscriptions

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -52,6 +52,7 @@ objects:
         values:
         - "true"
     resourceApplyMode: Sync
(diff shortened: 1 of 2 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -52,6 +52,7 @@ objects:
         values:
         - "true"
     resourceApplyMode: Sync
+    applyBehaviour: CreateOrUpdate
```
jimdaga, line 55: Should be `applyBehavior`

jimdaga (commented):
One small typo that needs fixing

PR #449, opened 2026-04-06
chore: migrate remaining SSS resources into cluster package

Ajpantuso:
### Summary

Adds back resources originally owned by the OLM SSS template.

### DIFF
--- a/deploy_pko/ClusterRoleBinding-cloud-ingress-operator.yaml
+++ b/deploy_pko/ClusterRoleBinding-cloud-ingress-operator.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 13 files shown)


abyrne55:
Hey @Ajpantuso, thanks for catching the missing resources here (e.g., the ClusterRoleBinding I missed with #447)

I've folded your changes into #447, which also covers the SSS namespace ordering fix, the SSS rename, and the cleanup Job immutability fix. One heads-up: the `credentials` phase on the CredentialsRequests here isn't defined in `manifest.yaml` (only `crds`, `namespace`, `rbac`, `deploy`, `cleanup-rbac`, `cleanup-deploy`), which I think might break things. I think I've addressed it in #447 by using `deploy`.

If you agree, could you close this and lgtm #447?

Ajpantuso:
Closing in favor of #447

PR #450, opened 2026-04-07
SREP-1550 Fix GCP e2e test flakiness in LB reconciliation

abyrne55:
## Summary

Ports the [OSD-28341](https://issues.redhat.com/browse/OSD-28341) AWS e2e fixes to the GCP code path, addressing the intermittent GCP LB reconciliation test failure documented in [SREP-1550](https://issues.redhat.com/browse/SREP-1550).

Two root causes:
- **CIDR test interference**: The `slices.Equal` polling compared the in-memory object to itself (always immediately true), so the operator could still be reconciling CIDR changes when the LB deletion test started. Fixed by re-fetching the actual Service and checking `LoadBalancerSourceRanges`. Moved CIDR restoration to a `defer` and used proper slice copies to avoid shared-array mutation.
- **GCP deletions not waiting for completion**: Each GCP `Delete` call returns an `Operation` that must be polled. The old code ignored these and moved to the next deletion immediately, causing dependent resources (e.g., target pool referenced by a still-deleting forwarding rule) to fail silently. Now waits via `RegionOperations.Wait`/`GlobalOperations.Wait` after each deletion and fails the test on errors.

## Test plan

- [x] `make go-build` passes
- [x] `make go-test` passes
- [ ] GCP e2e test passes on integration (hivei01ue1)

[OSD-28341]: https://redhat.atlassian.net/browse/OSD-28341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SREP-1550]: https://redhat.atlassian.net/browse/SREP-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/e2e/cloud_ingress_operator_tests.go
+++ b/test/e2e/cloud_ingress_operator_tests.go
@@ -10,7 +10,6 @@ import (
 	"log"
 	"net"
 	"os"
-	"slices"
 	"time"
 
 	cloudingressv1alpha1 "github.com/openshift/cloud-ingress-operator/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


joshbranham (commented):
/lgtm

PR #453, opened 2026-04-10
SREP-1550 Fix GCP address deletion using resource name instead of IP

abyrne55:
## Summary

Fixes the root cause of the persistent GCP e2e LB reconciliation test failure. Follow-up to #450 and #452.

`Addresses.Delete(project, region, name)` expects a **resource name**, but the test was passing the **IP address** (e.g. `34.75.221.55`), causing a GCP API 400 error. The error was silently logged, leaving the address reservation in place. When the operator recreated the rh-api Service, GCP assigned the same IP from the persisting reservation, so `newLBIP == oldLBIP` and the test waited forever.

Fix: list addresses in the region, find the one whose `.Address` matches the old LB IP, then delete by `.Name`.

## Test plan

- [ ] GCP e2e test passes on integration


joshbranham (commented):
/lgtm

PR #454, opened 2026-04-10
SREP-1550 Fix GCP e2e to check forwarding rule, not service IP

abyrne55:
## Summary

Fixes the GCP LB reconciliation e2e test that has been timing out on every run. Follow-up to #450, #452, #453.

**Root cause:** The test waited for the rh-api Kubernetes Service to get a new IP after deleting its backing GCP LB resources. This can never work because:
1. The cloud-ingress-operator doesn't delete/recreate the Service — it only ensures it exists
2. The cloud-controller-manager recreates the GCP LB behind the existing Service, reusing the same IP
3. So `newLBIP == oldLBIP` forever → 20-minute timeout

**Fix:** Instead of polling the Service IP, poll GCP directly for a new forwarding rule with a different name from the one we deleted. This checks what actually matters: did the platform recreate the load balancer.

Confirmed on live clusters `osde2e-6jyeo` and `osde2e-mmmtt` that the operator recreates the Service within ~2 min and the CCM recreates the forwarding rule with a new name, but reuses the same IP.

## Test plan

- [ ] GCP e2e test passes on integration

### DIFF
--- a/.tekton/cloud-ingress-operator-pko-pull-request.yaml
+++ b/.tekton/cloud-ingress-operator-pko-pull-request.yaml
@@ -13,7 +13,7 @@ metadata:
     appstudio.openshift.io/application: cloud-ingress-operator
     appstudio.openshift.io/component: cloud-ingress-operator-pko
     pipelines.appstudio.openshift.io/type: build
-  name: cloud-ingress-operator-pko-on-pull_request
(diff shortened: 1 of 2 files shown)


joshbranham (commented):
/lgtm

PR #456, opened 2026-04-17
Change old SSS to upsert.

bergmannf:
This way the namespace will not be terminated when moving to PKO.

## Summary by CodeRabbit

* **Chores**
  * Modified configuration application behavior to better preserve existing resource customizations during updates.


bergmannf:
Merged by hand because no owners in timezone.

